PR #16, opened 2021-05-25
Fix loop

Henryws:
codefactor says `Since you double quoted this, it will not word split, and the loop will only run once. `


Henryws:
codefactor was wrong

PR #18, opened 2021-05-25
Fix upgrade issue

Henryws:
What happened is pacstall upgrades in a way that doesn't work anymore and it wanted to upgrade packages like this: `sudo pacstall -I pkg1,pkg2,pkg3` which doesn't work


wizard-28 (approved):
Good to go but remember to merge this both into `master` and `develop`

PR #26, opened 2021-05-27
Fix bugs in the update code block

wizard-28:
This PR fixes bugs in the update code block in `pacstall` file


Henryws (approved):
looks good

PR #28, opened 2021-05-27
Fix showing blank dependencies when removing packages

wizard-28:
This PR fixes the issue of pacstall showing this warning when removing a package
```
You may want to remove:
```
Its showing `<blank>` as there are no dependencies to show, this PR fixes this


Henryws (approved):
looks good

PR #35, opened 2021-05-28
 [feature] Add filters to the labeler

wizard-28:
This adds filters to the labeler


wizard-28:
Okay testing it for the last time now

PR #36, opened 2021-05-28
[feature] Add filters to the labeler

wizard-28:
This adds filters to the labeler
-----------------
If this works corrently then it should ignore tags lile
help, hotfix, bug
in the description and only look for tags in the title


wizard-28:
Looks like it worked this time

PR #44, opened 2021-05-31
[bug] Remove `fzf` dependency

wizard-28:
It is a redundant dependency, as it is no longer used. This PR removes that dependency from `install.sh`


Henryws (approved):
Looks good

PR #45, opened 2021-06-01
Add tree feature with `-T`

Henryws:


wizard-28 (changes requested):
Add `tree` as a pacstall dependency during install, otherwise bash won't find the `tree` command!

wizard-28 (approved):
Good to go, remember to squash merge!

PR #50, opened 2021-06-02
[Feature] Add stow install if not found

Henryws:
This is when the installer fails to install stow, so pacstall will install it if not found


wizard-28:
But doesn't it already install `stow` as a dependency...

wizard-28:
And please use the correct titling scheme.

Henryws:
> But doesn't it already install `stow` as a dependency...

It is but people have problems with it so better safe than github issues

wizard-28 (approved):
Okay then!

PR #53, opened 2021-06-06
[bug] Package should be removed if the grep matches

kaistian:


misc/scripts/install-local.sh
```diff
@@ -126,7 +126,7 @@ if echo -n "$depends" > /dev/null 2>&1; then
 fi
 if ! [[ -z $replace ]] ; then
```
wizard-28, line 127:
```suggestion
if [[ -n $replace ]] ; then
```

misc/scripts/install-local.sh
```diff
@@ -126,7 +126,7 @@ if echo -n "$depends" > /dev/null 2>&1; then
...
     dpkg-query -W -f='${Status}' $replace 2>/dev/null | grep -q "ok installed"
-    if [[ $? -eq 1 ]] ; then
+    if [[ $? -eq 0 ]] ; then
```
wizard-28, line 129:
```suggestion
    if [[ $(dpkg-query -W -f='${Status}' $replace 2>/dev/null | grep -q "ok installed") -eq 0 ]] ; then
```

wizard-28 (changes requested):
Also, fix the CodeFactor issue

Henryws (approved):
looks good

wizard-28 (approved):
I'll merge

PR #57, opened 2021-06-08
[bug] Remove obsolete `rm`

wizard-28:
Says what it does, no need to `rm` the file, as `touch` already does that.


Henryws:
touch updates a files date properties, I want to remove and add again, not update the time properties

wizard-28:
Okay then I am adding ` > /dev/null` to it, to prevent warning when the file doesn't exist (the file doesn't exist when pacstall is installed for the first time)

Henryws:
but that's what `rm -f` does

wizard-28:
Then what is this?
image

Henryws:
hmm, I guess redirecting to `/dev/null` would be fine

Henryws (approved):
good

PR #59, opened 2021-06-08
Fixed some weird grammar

aranym0:
Nothing major, but since it's the first thing people will see, thought I might as well fix it


Henryws:
looks good

PR #70, opened 2021-06-13
Add basic explicitly installed thingy

Henryws:
This will export a variable that is only called when something is installing with pacdeps. The metadata logger will see that and log it as `installed as dependency`. If `-Qi` can't find the `$pacstall_depends`, it will print that is was explicitly installed. I'm testing now with `sudo pacstall -I soundux`

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -226,7 +226,7 @@ if [[ -n $build_depends ]]; then
     echo "build_dependencies=\"$build_depends"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
 fi
 if [[ -n $pacdeps ]]; then
-    echo "pacdeps=\"$pacdeps"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
+    echo "_pacdeps=\"$pacdeps"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
 fi
 if test -f /tmp/pacstall-pacdeps; then
     echo "pacstall_depends=\"true"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -228,7 +228,7 @@ if [[ -n $build_depends ]]; then
     echo "build_dependencies=\"$build_depends"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
 fi
 if [[ -n $pacdepslist ]]; then
-    echo "pacdeps=\"${pacdepslist[*]}"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
+    echo "pacdeps=\"$pacdepslist"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
 fi
 if [[ -n $PACSTALL_DEPENDS ]]; then
     echo "pacstall_depends=\"true"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -103,9 +103,9 @@ if [[ -n "$pacdeps" ]]; then
     for i in "${pacdeps[@]}"
     do
         fancy_message info "Installing $i"
-        sudo touch /tmp/pacstall-pacdeps
+        sudo touch /tmp/pacstall-pacdeps-$i
         sudo pacstall -P -I "$i"
-        sudo rm -f /tmp/pacstall-pacdeps
+        sudo rm -f /tmp/pacstall-pacdeps-$i
     done
 fi
 
@@ -228,7 +228,7 @@ fi
 if [[ -n $pacdeps ]]; then
     echo "_pacdeps=\"$pacdeps"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
 fi
-if test -f /tmp/pacstall-pacdeps; then
+if test -f /tmp/pacstall-pacdeps-$PACKAGE; then
     echo "_pacstall_depends=\"true"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
 fi
 
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -104,9 +104,11 @@ if [[ -n "$pacdeps" ]]; then
     for i in "${pacdeps[@]}"
     do
         fancy_message info "Installing $i"
+        echo "PACSTALL_DEPENDS=true" | sudo tee /tmp/pacstall-pacdeps
         export PACSTALL_DEPENDS=true
         sudo pacstall -P -I "$i"
         pacdepslist+=($i)
+        sudo rm -f /tmp/pacstall-pacdeps
         unset PACSTALL_DEPENDS
     done
 fi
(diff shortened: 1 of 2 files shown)


Henryws:
Doesn't work yet. Sigh

misc/scripts/install-local.sh
```diff
@@ -104,8 +104,10 @@ if [[ -n "$pacdeps" ]]; then
     for i in "${pacdeps[@]}"
     do
         fancy_message info "Installing $i"
+        echo "PACSTALL_DEPENDS=true" | sudo tee /tmp/pacstall-pacdeps
         sudo pacstall -P -I "$i"
         pacdepslist+=($i)
+        sudo rm -f /tmp/pacstall-pacdeps
```
wizard-28, line 110: You are writing to the file, and removing the in the same code?

Henryws: After installation of a package so that the metadata logger can log it as a dependency, then delete it

wizard-28: IF the logger logs it, and then deletes it, whats the use? Ultimately there is no log of it left.

Henryws: ? It's to check if a package is installed as a dependency. For instance, running `sudo pacstall -I soundux` will install `pipewire` as a dependency. So it makes the file, the logger detects it and shows that is was installed as a dependency when running `pacstall -Qi pipewire`. Then it deletes the file

wizard-28: This part of the code is logging it (writing it to the file), the detecting part starts from L233, so if the logger itself delets the file, then how will the detector (which comes later) detect it?

Henryws: Because running `sudo pacstall -I pkg` will at the end, look for the file that signifies that the package was installed as a dependency. Then it will delete after `sudo pacstall -I pkg` is done. It's running two pacstall instances

wizard-28: Omg, well it got way more complex.

wizard-28:
Okay so what are the different uses of "pacdeps", "pacdepslist" and "pacstall_depends", and can you please refactor these variable names to something that makes more sense?

Henryws:
`pacdeps` is the `pkg.pacscript` array. `pacdepslist` is redundant, now that I look at it. `pacstall_depends` is the log variable that `pacstall -Qi pkg` will look for and will print the pacstall packages needed to build it, like `pipewire` would be listed when running `pacstall -Qi soundux`, as `pipewire` was needed to build `soundux`

wizard-28:
Can you change all the log variable names to start with `_`

Like `_pacstall_depends` or just make it `_pacdeps`, so it is easier for new contributors to decipher the code.

Henryws:
`pacdeps` is a pacscript variable, so nothing can be changed there

misc/scripts/install-local.sh
```diff
@@ -225,8 +225,11 @@ fi
 if [[ -n $build_depends ]]; then
     echo "build_dependencies=\"$build_depends"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
 fi
-if [[ -n $pacdepslist ]]; then
-    echo "pacdeps=\"${pacdepslist[*]}"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
+if [[ -n $pacdeps ]]; then
+    echo "pacdeps=\"$pacdeps"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
+fi
```
wizard-28, line 230:
```suggestion
if [[ -n $pacdeps ]]; then
    echo "_pacdeps=\"$pacdeps"\" | sudo tee -a /var/log/pacstall_installed/"$PACKAGE" >/dev/null
fi
```
If I understand correctly this portion's `pacdeps` is the pacscript pacdeps, which is fine, What I am asking you to do is write it to the file using the `_` prefix.

pacstall
```diff
@@ -394,15 +394,20 @@ while [[ "$1" =~ ^- && ! "$1" == "--" ]]; do
       if [[ -n $maintainer ]]; then
         echo "maintainer: ""$maintainer"""
       fi
-      if [[ -n $pacdepslist ]]; then
-          echo "pacstall dependencies: ""${pacdepslist[*]}"""
+      if [[ -n $pacdeps ]]; then
+          echo "pacstall dependencies: ""$pacdeps"""
```
wizard-28, line 398:
```suggestion
      if [[ -n $_pacdeps ]]; then
          echo "pacstall dependencies: ""$_pacdeps"""
```

Then read it here with the prefix

misc/scripts/install-local.sh
Henryws: oh

wizard-28:
Do that for all the logging variables.

That way it is easier for people to understand which variables are used for logging and which are from the pacscript

pacstall
```diff
@@ -387,21 +387,26 @@ while [[ "$1" =~ ^- && ! "$1" == "--" ]]; do
...
       source /var/log/pacstall_installed/"$PACKAGE"
       echo "name: $PACKAGE"
-      echo "version: $version"
+      echo "version: $_version"
       echo "size: $(du -sh /usr/src/pacstall/"$PACKAGE" | awk '{print $1}')"
-      echo "description: ""$description"""
-      echo "date installed: ""$date"""
+      echo "description: ""$_description"""
+      echo "date installed: ""$_date"""
       if [[ -n $maintainer ]]; then
```
wizard-28, line 394:
```suggestion
      if [[ -n $_maintainer ]]; then
```

pacstall
```diff
@@ -387,21 +387,26 @@ while [[ "$1" =~ ^- && ! "$1" == "--" ]]; do
...
-        echo "maintainer: ""$maintainer"""
+        echo "maintainer: ""$_maintainer"""
       fi
-      if [[ -n $pacdepslist ]]; then
-          echo "pacstall dependencies: ""${pacdepslist[*]}"""
+      if [[ -n $_pacdeps ]]; then
+          echo "pacstall dependencies: ""$_pacdeps"""
       fi
       if [[ -n $dependencies ]]; then
```
wizard-28, line 400:
```suggestion
      if [[ -n $_dependencies ]]; then
```

pacstall
```diff
@@ -387,21 +387,26 @@ while [[ "$1" =~ ^- && ! "$1" == "--" ]]; do
...
-          echo "dependencies: ""$dependencies"""
+          echo "dependencies: ""$_dependencies"""
       fi
       if [[ -n $build_dependencies ]]; then
```
wizard-28, line 403:
```suggestion
      if [[ -n $_build_dependencies ]]; then
```

pacstall
```diff
@@ -387,21 +387,26 @@ while [[ "$1" =~ ^- && ! "$1" == "--" ]]; do
...
-          echo "build dependencies: ""$build_dependencies"""
+          echo "build dependencies: ""$_build_dependencies"""
+      fi
+      if [[ -n $pacstall_depends ]]; then
```
wizard-28, line 406:
```suggestion
      if [[ -n $_pacstall_depends ]]; then
```

wizard-28 (changes requested):

pacstall
Henryws: forgot those, sorry

misc/scripts/install-local.sh
```diff
@@ -100,12 +100,12 @@ if [[ -n "$build_depends" ]]; then
     fi
 
 if [[ -n "$pacdeps" ]]; then
-    declare -a pacdepslist
     for i in "${pacdeps[@]}"
     do
         fancy_message info "Installing $i"
+        echo "PACSTALL_DEPENDS=true" | sudo tee /tmp/pacstall-pacdeps
```
wizard-28, line 106:
```suggestion
        sudo touch /tmp/pacstall-pacdeps
```
There is no need of echoing anything there, I think, because in the detector, you are just checking if the file exists

Henryws:
merge?

wizard-28:
Have you tested it with recursive pacdeps?

I mean A installs pacdeps B, but B installs pacdeps C?

Henryws:
Hm, good point, maybe the script could look for `/tmp/pacstall-pacdeps-$PACKAGE` instead of `/tmp/pacstall-pacdeps`

wizard-28:
Okay, just test it once in your computer, with a recursive pacscript, then I'll merge

Henryws:
I will need to push a dummy pacscript to the repo to test though

wizard-28:
> I will need to push a dummy pacscript to the repo to test though

Do a `-Il`

Henryws:
But if it has a `pacdeps` array, it will pull from the repo

wizard-28:
Fix the Codacy warnings

PR #74, opened 2021-06-15
Fix `-R` showing file error

Henryws:
When removing a package that doesn't exist, this shows:
```bash
➜  ~ sudo pacstall -R soundux
/bin/pacstall: line 243: /var/log/pacstall_installed/soundux: No such file or directory
[!] ERROR: soundux does not exist or not properly symlinked
```
This pr fixes the no such file or directory


Henryws:
Merge?

wizard-28 (approved):

PR #78, opened 2021-06-17
Fix: Phantom Updates

wizard-28:


wizard-28:
And as a sidenote, why are the logs all over the place, firstly, some are there in `/tmp/pacstall`, one other is in `/tmp/pacstall-up-list`, other is in `/tmp/pacstall-optdepends`?

Henryws (approved):
Looks good

Henryws:
> And as a sidenote, why are the logs all over the place, firstly, some are there in `/tmp/pacstall`, one other is in `/tmp/pacstall-up-list`, other is in `/tmp/pacstall-optdepends`?

Because on is a list for upgrades, and another is for optional dependencies...

wizard-28:
I meant where aren't they all in the `/tmp/pacstall/` directory?

Henryws:
> I meant why aren't they all in the `/tmp/pacstall/` directory?

Because they would get deleted at possibly the wrong time

PR #81, opened 2021-06-17
Feat: Add global `*dir` constants

Henryws:
`srcdir` is the package's source directory. For example, pkg foo has a
directory that is extracted to `/tmp/pacstall` called `foo-1.0`. The
`srcdir` would be `/tmp/pacstall/foo-1.0`

`pkgdir` is the path that a package is to be installed to. Package
`foo`'s `pkgdir` would be `/usr/src/pacstall/foo`


wizard-28:
Can you add `homedir` too, It should point to `/home/$(logname)`

wizard-28:
Okay, mark this PR ready for review when you like it.

wizard-28 (changes requested):
Can you remove the `/` from `srcdir`?
img

Issue #77, opened 2021-06-09
Add `.deb` support in pacstall


NarutoXY:
Can we include .deb packages for some packages like Brave Nightly in Pacstall? It will help many users who don't want to just compile packages like a web browser. It will make it even more better and similar to AUR where both git and binaries are available.

emanuelserpa:
I think pacstall could just build debs.

I'm not using Ubuntu currently, but building debs (https://linuxconfig.org/easy-way-to-create-a-debian-package-and-local-package-repository) doesn't look that hard.

NarutoXY:
I created pull request regarding .deb support but it is conflicting due to some recent changes. https://github.com/pacstall/pacstall/pull/93

PR #93, opened 2021-06-25
Feat: Added Alpha `.deb` support using `wget` and `awk`

NarutoXY:
Hey guys! I edited install-local script to support .deb packages, using wget and awk.

Related issues: #77

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -194,7 +194,11 @@ else
   elif [[ $url = *.deb ]]; then
   	hashcheck "${url##*/}"
   	cd $(/bin/ls -d -- */|head -n 1)
+  	fancy_message info "Installing"
   	sudo apt install -f $(echo "$url" | awk -F "/" '{print $NF}')
+  	echo "_version=\"$version"\" | sudo tee "$LOGDIR"/"$PACKAGE" > /dev/null
+  	echo "_description=\"$description"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
+  	echo "_date=\"$(date)"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
   	exit 0
 
   else
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -193,7 +193,6 @@ else
 
   elif [[ $url = *.deb ]]; then
   	hashcheck "${url##*/}"
-  	cd $(/bin/ls -d -- */|head -n 1)
   	fancy_message info "Installing"
   	sudo apt install -f $(echo "$url" | awk -F "/" '{print $NF}')
   	echo "_version=\"$version"\" | sudo tee "$LOGDIR"/"$PACKAGE" > /dev/null
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -60,7 +60,7 @@ function cget() {
 }
 
 # logging metadata
-function logging_meta() {
+function loggingMeta() {
 	echo "_version=\"$version"\" | sudo tee "$LOGDIR"/"$PACKAGE" > /dev/null
 	echo "_description=\"$description"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
 	echo "_date=\"$(date)"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
@@ -71,10 +71,10 @@ function logging_meta() {
 }
 
 # function to install .deb packages and check exit code
-funtion install_deb_packages() {
+funtion installDebPackages() {
     sudo apt install -f $(echo "$url" | awk -F "/" '{print $NF}') 2>/dev/null
     if [[ $? -eq 0 ]]; then
-    	logging_meta
+    	loggingMeta
     else
     	fancy_message error "Failed to install the package"
     fi
@@ -235,7 +235,7 @@ else
   elif [[ $url = *.deb ]]; then
     hashcheck "${url##*/}"
     fancy_message info "Installing"
-    install_deb_packages
+    installDebPackages
   else
     # I think you get it by now
     hashcheck "${url##*/}"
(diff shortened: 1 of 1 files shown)


Henryws:
How would updating work?

NarutoXY:
Oh I would look further into updating the packages. I have a fix that we see changes in the version (mostly all deb packages have their version and a uniform way to name the package) and then we can wget the new package and apt install it. I will push update related changes by tomorrow 17:00 GMT

NarutoXY:
Most deb packages and online repository while installation which takes care of the update process and get updated during the system upgrade

NarutoXY:
I just looked at the update script. If I'm correct then, pacstall will compare the pkgver and do pacstall -I $package to update the packages, which in turn will install the new .deb package and apt will install the package (since application will be same). So AFAIK there's no need for any extra update script.

NarutoXY:
Hopefully I have resolved the conflicts

Henryws:
@NarutoXY Have you actually tested this yet?

NarutoXY:
I haven't tested this 😅 on real hardware. My machine got destroyed last week and everything was needed to be wiped. So I'm currently setting up my machine. 

But I believe that this will work as I have reviewed it multiple times and ran it in VM.

NarutoXY:
Fired a live Ubuntu ISO, installed brave deb package using pacstall. So, it should work

Henryws:
Could I see the pacscript?

NarutoXY:
I just created one in hurry for test and it worked. 

```name="brave-browser-beta"
pkgname="brave-browser-beta"
version="1.27.81"
url="https://github.com/brave/brave-browser/releases/download/v1.27.81/brave-browser-beta_1.27.81_amd64.deb"
build_depends=""
depends=""
gives="brave-browser-beta"
description="Brave Deb file for test (Alpha stage)"
breaks=""```

**Comment by Henryws** [comment]
> I just created one in hurry for test and it worked.
> 
> ```
> name="brave-browser-beta"
> pkgname="brave-browser-beta"
> version="1.27.81"
> url="https://github.com/brave/brave-browser/releases/download/v1.27.81/brave-browser-beta_1.27.81_amd64.deb"
> build_depends=""
> depends=""
> gives="brave-browser-beta"
> description="Brave Deb file for test (Alpha stage)"
> breaks=""
> ```
> 
> I just created this dummy pacscript for test, and it worked. But for the real pacscript (to be on GitHub) I need to use variables like $version in place of 1.27.81 in url

It needs a hash to work

**Comment by NarutoXY** [comment]
I'm busy in classes. Could you please just type in that hash value?

**Comment by Henryws** [comment]
Can you not use '&&' in the logging part?

**Comment by NarutoXY** [comment]
I just used && but I can't understand what you mean. Please share the snippet or edit it.

**Comment by Henryws** [comment]
Line 197

**Comment by NarutoXY** [comment]
Ook, but I've used ampersands to prevent any mishap when apt fails to install. So no logging is done when apt fails and error is printed. If I would have to remove them then I've to create one more if stat

**Comment by Henryws** [comment]
> Ook, but I've used ampersands to prevent any mishap when apt fails to install. So no logging is done when apt fails and error is printed. If I would have to remove them then I've to create one more if statement to check exit code of apt

Can you do that instead

**Comment by NarutoXY** [comment]
Yes I can but I will be busy in completing my school homework (literally 500 pages). So I'll contribute after 6 July probably.

**Comment by Henryws** [comment]
> Yes I can but I will be busy in completing my school homework (literally 500 pages). So I'll contribute after 6 July probably. 

Ok

**Comment by Henryws** [comment]
On the functions part, if it's only used once, why make it a function?

**Comment by NarutoXY** [comment]
Because ```nested if``` is not supported in shell scripts and it will cause many errors. So I just wrapped everything in a function. In this way, we can even log more details about the package installed.

**Comment by Henryws** [comment]
So is this ready to merge?

**Comment by NarutoXY** [comment]
On my end, it's complete. But, as precaution, you guys should test it on your machines too.

**Comment by Henryws** [comment]
> On my end, it's complete. But, as precaution, you guys should test it on your machines too. 

image

**Comment by NarutoXY** [comment]
Try now, maybe it was with ```_``` in function name.

**Comment by Henryws** [comment]
Actually I'm going boating for 4 hours so could you test it?

**Review by Henryws — APPROVED** [review]
Looks good, merging

**Comment by kaistian** [comment]
> Because `nested if` is not supported in shell scripts and it will cause many errors.

Where to you have that from?
Pacstall have several nested if and don't have any problem with it and it's even documented for bash https://tldp.org/LDP/abs/html/nestedifthen.html

**Comment by NarutoXY** [comment]
Hey I mean nested if don't work when we have many different conditions to check. We are checking file type,then exit code then the variables for logging and this can't work.

**Comment by kaistian** [comment]
There would be no conflict and nothing to figure out, just paste the code in instead of the function name like this
```
  elif [[ $url = *.deb ]]; then
    hashcheck "${url##*/}"
    fancy_message info "Installing"
    sudo apt-get install -f $(echo "$url" | awk -F "/" '{print $NF}') 2>/dev/null
    if [[ $? -eq 0 ]]; then
      echo "_version=\"$version"\" | sudo tee "$LOGDIR"/"$PACKAGE" > /dev/null
      echo "_description=\"$description"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
      echo "_date=\"$(date)"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
      if [[ $removescript == "yes" ]]; then
        echo "_removescript=\"yes"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
      fi
      echo "_maintainer=\"$maintainer"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
    else
      fancy_message error "Failed to install the package"
    fi
  else
```
This is the same code as yours without functions.

That said the code doesn't work because
* apt will try to install a package and not the file, for that you need to provide absolute or a relative path.
* The meta date will be overwritten further down the code https://github.com/pacstall/pacstall/blob/6ff21429617842c8871a88fe080a7243f4b00cc7/misc/scripts/install-local.sh#L282
* It error on stow, since there are nothing to stow https://github.com/pacstall/pacstall/blob/6ff21429617842c8871a88fe080a7243f4b00cc7/misc/scripts/install-local.sh#L328

Henryws:
I'll add an if statement detecting the url ending so it won't stow debs

NarutoXY:
Pacstall now supports ```.deb``` files. https://github.com/pacstall/pacstall/pull/115

PR #106, opened 2021-07-11
Fix: /root is not /home/root

Henryws:
So a normal user has a `/home` directory, but what about root? Root's home directory is `/root`. This fixes that

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -24,7 +24,6 @@
 
 # Configuration
 export LC_ALL=C
-export HOMEDIR="/home/$(logname)"
 export LOGDIR="/var/log/pacstall"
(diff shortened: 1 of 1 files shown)


pacstall
```diff
@@ -27,7 +27,7 @@ export LC_ALL=C
 export HOMEDIR="/home/$(logname)"
```
wizard-28, line 27: I think we should remove `HOMEDIR` too, since pacstall is no longer running under root, so people can directly use `~` to access user's home directory.

Henryws: true

wizard-28 (approved):

PR #110, opened 2021-07-11
Fix: pacstall tried to create /.cache

Henryws:


misc/scripts/download.sh
```diff
@@ -33,7 +33,7 @@ if ! wget -q --tries=10 --timeout=20 --spider https://github.com; then
 fi
 
 if curl --output /dev/null --silent --head --fail "$URL" ; then
-  mkdir -p "$HOMEDIR/.cache/pacstall" && cd "$HOMEDIR"/.cache/pacstall/
+  mkdir -p "$HOME/.cache/pacstall" && cd "$HOME"/.cache/pacstall/
```
wizard-28, line 36: why not use `~` instead of `$HOME`?

PR #113, opened 2021-07-11
Fix: install script won't install files with root

Henryws:


kaistian:
You left out the `sudo mkdir`

Henryws:
oh, just missed it

PR #115, opened 2021-07-12
 Fix typos

NarutoXY:


misc/scripts/install-local.sh
```diff
@@ -59,6 +59,30 @@ function cget() {
...
+	if [[ $removescript == "yes" ]]; then
+	  echo "_removescript=\"yes"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
+	fi
+	echo "_maintainer=\"$maintainer"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
+}
+
+
+#Function to install .deb files
+function debpt() {
+    sudo apt install -f "$(echo "$url" | awk -F "/" '{print $NF}')"
```
kaistian, line 76: Install is something that is done in the install function of the pacscript so this would break that assumption.

Also the apt install wont work. Lets say `url="https://github.com/lbryio/lbry-desktop/releases/download/v0.51.1/LBRY_0.51.1.deb"` it will then try to run `sudo apt install -f LBRY_0.51.1.deb` from the repository. To make it install from a file you need to specify a path.

misc/scripts/install-local.sh
```diff
@@ -211,6 +235,10 @@ else
     # Make the directory available for users
     sudo chown -R "$(logname)":"$(logname)" .
 
+  elif [[ $url = *.deb ]]; then
+    hashcheck "${url##*/}"
+    fancy_message info "Installing"
```
kaistian, line 240: This will produce to messages about installing.
The install part is the function install on line 246-247.

kaistian (commented):
This way of installing a deb will different from already practice of using the install function.
If this is OK other part should be rewritten like the install function so that is doesn't run.

misc/scripts/install-local.sh
NarutoXY: Please see clearly I've specified exit codes, so it exists after install

kaistian: I did not see the exit and it was impossible for me to check since it fails when running it
```
❯ pacstall -Il brave-browser-beta 
Do you want to view the pacscript first [y/N] 
Do you want to edit the pacscript [y/N] 
[+] INFO: Sourcing pacscript
[+] INFO: URL exists
[+] INFO: Installing dependencies
[+] INFO: Retrieving packages
brave-browser-beta_1.27.100_amd64.deb                          100%[====================================================================================================================================================>]  82.85M  28.4MB/s    in 2.9s    
[+] INFO: Installing
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package brave-browser-beta_1.27.100_amd64.deb
[!] ERROR: Failed to install the package
```

There is a huge problem with you exit, later in the code pacstall does cleanup, since you exit to early it will never run.

NarutoXY: Hey could you give a possible fix or explain problem in detail. I think we can use ```pwd``` to get full path and it would solve the problem

misc/scripts/install-local.sh
```diff
@@ -59,6 +59,30 @@ function cget() {
...
+# logging metadata
+function loggingMeta() {
+	echo "_version=\"$version"\" | sudo tee "$LOGDIR"/"$PACKAGE" > /dev/null
+	echo "_description=\"$description"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
+	echo "_date=\"$(date)"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
+	if [[ $removescript == "yes" ]]; then
+	  echo "_removescript=\"yes"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
+	fi
+	echo "_maintainer=\"$maintainer"\" | sudo tee -a "$LOGDIR"/"$PACKAGE" > /dev/null
+}
```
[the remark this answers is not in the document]
NarutoXY: We use exit code so it doesn't get overwritten

misc/scripts/install-local.sh
NarutoXY: apt also accepts relative paths

NarutoXY: We can also use pwd to solve this

misc/scripts/install-local.sh
kaistian: It won't solve the clean up since you exit before it run, but the install part can be fixed by adding ./ in front `sudo apt install -f "./$(echo "$url" | awk -F "/" '{print $NF}')"`

A much better solution is to drop the install part at let that be in the pacscript, then install and meta will work that already is there. And the cleanup will also be run.

The pacscript will then look like this
```
name="brave-browser-beta"
pkgname="brave-browser-beta"
version="1.27.100"
url="https://github.com/brave/brave-browser/releases/download/v$version/brave-browser-beta_${version}_amd64.deb"
build_depends=""
depends=""
gives="brave-browser-beta"
description="Brave Deb file for test (Alpha stage)"
breaks=""
hash="4d7d98048273bf05877d6303efe94a2c9ad01d4fe4ef05c1b2cfd26033dcf9e4"

prepare() {
        true
}

build() {
        true
}

install() {
        sudo apt-get install "./brave-browser-beta_${version}_amd64.deb"
}
```

The benefits with is is all the functions in pacstall remain operational.

NarutoXY: But the current philosophy also does the trick and tar would complain that it isn't archive and would exit out, that could be solved using empty case statement

NarutoXY: @kaistian try running it now, Twilight blood applied some changes

kaistian: Well it doesn't clean up after it self so I would say the philosophy is a little broken.
Try it and check /tmp/pacstall

kaistian:
What you are doing now is duplicate the code over and over again, that is not good practice and will become error and bug prone since a bug will be fixed one place but forgotten every place the code is duplicated.
It also make is hard to maintain the code and for new people to understand what is actually happening.

NarutoXY:
No we can just use loggingMeta both the places and then no code reperition. I didn't do it at first place as I was a little bit skeptical.

PR #116, opened 2021-07-12
```case``` instead of ```if```

NarutoXY:
Done what I promised, shellcheck has no issues.


NarutoXY:
@Henryws it looks good to me

PR #117, opened 2021-07-12
The remove doesn't work

kaistian:
The if is very strange and it will always run and the checking for removescript is done before the sourcing of the file.
Stow is missing --target so it doesn't remove any of the symlink.

### DIFF
--- a/misc/scripts/remove.sh
+++ b/misc/scripts/remove.sh
@@ -21,7 +21,7 @@ fi
 source "$LOGDIR/$PACKAGE" > /dev/null 2>&1
 
 fancy_message info "Removing symlinks"
-sudo stow -D "$PACKAGE"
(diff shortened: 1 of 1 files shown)


Henryws:
Have you tested the `fn_exists` to make sure it still works without returning anything?

kaistian:
Yes, testet with and without removescript in pacsript, removescript was just true.
```
$ pacstall -R bat
[+] INFO: Removing symlinks
[+] INFO: Removing package
[+] INFO: Running post removal script
```
```
$ pacstall -R bat                                   
[+] INFO: Removing symlinks
[+] INFO: Removing package
```

Henryws:
You mean the removescript variable or the function?

kaistian:
Function.

PR #129, opened 2021-07-18
Hotfixes for v1.5.1

wizard-28:
I intend to commit hotfixes for these problems.

- [x] `STGDIR` in installer pointing to the root user's `.local` directory (should be to the normal user's)
- [x] Completions ( `-I` flag mainly) for the `fish`, and `bash` shells are not working. (This is a temporary fix. @D-Brox is supposed to work on a more flexible permanent fix)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -96,7 +96,7 @@ sudo apt-get -qq update
 fancy_message info "Installing packages"
 sudo apt-get install -qq -y {curl,wget,stow,build-essential,unzip,tree,dialog}
 
-export STGDIR="${HOME}/.local/share/pacstall"
+export STGDIR="${LOGNAME}/.local/share/pacstall"
 fancy_message info "making directories"
 mkdir -p $STGDIR
 mkdir -p $STGDIR/scripts
--- a/install.sh
+++ b/install.sh
@@ -96,13 +96,9 @@ sudo apt-get -qq update
 fancy_message info "Installing packages"
 sudo apt-get install -qq -y {curl,wget,stow,build-essential,unzip,tree,dialog}
 
-case "$LOGNAME" in
-  "root")
-    export STGDIR="/usr/share/pacstall"
-    ;;
-  *)
-    export STGDIR="/home/${LOGNAME}/.local/share/pacstall"
-esac
+
+export STGDIR="/usr/share/pacstall"
+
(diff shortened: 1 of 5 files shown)


NarutoXY (changes requested):
```
case "$LOGNAME" in
"root")
    export STGDIR="/usr/share/pacstall"
;;
*)
    export STGDIR="${LOGNAME}/.local/share/pacstall"
;;
esac
```
This would use /usr/share in case using with root

install.sh
```diff
@@ -96,7 +96,13 @@ sudo apt-get -qq update
 fancy_message info "Installing packages"
 sudo apt-get install -qq -y {curl,wget,stow,build-essential,unzip,tree,dialog}
```
NarutoXY, line 97:
```
sudo apt-get install -qq -y {curl,wget,stow,build-essential,unzip,tree,dialog} 2>/dev/null
```

This would suppress warning of apt regarding stable CLI interface

NarutoXY (approved):

install.sh
wizard-28: It's using `apt-get` not `apt`, so there will be no warnings printed.

NarutoXY: See output of remote bash in tests.

NarutoXY: I misunderstood the base install part as pacstall installer

install.sh
```diff
@@ -96,7 +96,13 @@ sudo apt-get -qq update
...
 
-export STGDIR="${LOGNAME}/.local/share/pacstall"
+case "$LOGNAME" in
+  "root")
+    export STGDIR="/usr/share/pacstall"
+    ;;
+  *)
+    export STGDIR="/home/${LOGNAME}/.local/share/pacstall"
```
NarutoXY, line 104: I think it needs ```;;``` here before ```esac```

wizard-28: It doesn't need it, as it's the `default`/`last` case, no chance of fall through i suppose.

Henryws (approved):
👍🏻

Henryws (approved):
👍

NarutoXY (approved):

NarutoXY:
Should we merge now??

D-Brox (approved):

PR #133, opened 2021-07-19
Create CODE_OF_CONDUCT.md

NarutoXY:


Henryws (commented):
No, we don't need a code of conduct. I expect everyone to use common sense when communicating with others and I don't think a legally binding document is the best way to do it, so no, no code of conduct

PR #135, opened 2021-07-19
Feat: Replace `aria2` with `axel` for faster speeds

NarutoXY:
Go crazy.

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -92,7 +92,7 @@ function loggingMeta() {
 
 
 function aria2 {
-Parallel=$nproc + 5
+Parallel=$(($nproc + 5))
 if [[ -f "$STGDIR/repo/pacstallrepo.txt" ]]; then
   source "$STGDIR/repo/pacstallrepo.txt"
 fi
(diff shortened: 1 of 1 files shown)


misc/scripts/install-local.sh
```diff
@@ -92,9 +92,13 @@ function loggingMeta() {
 
 
 function aria2 {
+Parallel=$nproc + 5
```
D-Brox, line 95: I think it's best to use `bc` for safe string operation

D-Brox: Codefactor has a better option

NarutoXY: May you edit this?? I'm not familiar with bc

D-Brox: Codefactor had a better option, actually

NarutoXY: Ok, did changes

wizard-28:
Hey what about `-j`?
img

D-Brox:
@wizard-28  `-j` is for concurrent downloads, so multiple files. We can use it on `install.sh` and `update.sh` for downloading all scripts simultaneously, tho

NarutoXY:
> Hey what about `-j`?
> img

Have mercy on CPUs and ```-j``` will be used in install.sh and update.sh as @D-Brox pointed out its for multi file download

PR #136, opened 2021-07-20
Fix: exit 1 when asked N

NarutoXY:
Hey, may this fix the issue


pacstall
```diff
@@ -144,10 +144,18 @@ function ask() {
     fi
 
     # Check if the reply is valid
+
```
D-Brox, line 147: It's missing a check for `no default`
`if ask "question"` enters an infinite loop

D-Brox (changes requested):

PR #140, opened 2021-07-21
Fix: store pacscripts and use ```axel```

NarutoXY:
# Purpose
This PR makes use of `axel` to *axel*elerate downloads, instead of `aria2c` as it proved slower and doesn't have a progress bar.

This PR also fixes issue of packages not being removed and implements a new file for ```test``` workflow. 

## Progress Report
- [x] Use `axel` in `install-local.sh`
- [x] Use `axel` to download `pacscripts`
- [ ] Use `axel` in `update.sh`
- [ ] Use `axel` in the installer
- [ ] Fix removal of packages
- [x] New script for workflow

### DIFF
--- a/.github/workflows/remote-bash.yml
+++ b/.github/workflows/remote-bash.yml
@@ -9,4 +9,4 @@ jobs:
         with:
           args:
         env:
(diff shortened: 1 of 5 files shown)


wizard-28:
Hey, I was supposed to do this...

NarutoXY:
> Hey, I was supposed to do this...

You were but now I did

PR #152, opened 2021-07-23
Feat: Add fakeroot

Henryws:
# Purpose

This pr will make packages prepare and build inside of fakeroot, allowing for better security

# Approach

Pacstall will export all possible variables and functions a pacscript can have (obviously some will error, but as long as every variable/function that is in the script is exported, it's fine). Then it can pass into fakeroot

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -143,7 +143,7 @@ if [[ "$answer" -eq 1 ]]; then
     sudo apt-get install -qq -y ripgrep
 fi
 
-sudo apt-get install -qq -y {curl,wget,stow,build-essential,unzip,tree,dialog,bc}
+sudo apt-get install -qq -y curl wget stow build-essential unzip tree dialog bc fakeroot
 
 
 export STGDIR="/usr/share/pacstall"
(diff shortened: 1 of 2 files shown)


Henryws:
Currently the variables/functions all export (with a couple errors (but expected ones), but fakeroot doesn't think `build` and `prepare` are functions (which running from a terminal one step at a time does work)

install.sh
```diff
@@ -143,7 +143,7 @@ if [[ "$answer" -eq 1 ]]; then
     sudo apt-get install -qq -y ripgrep
 fi
 
-sudo apt-get install -qq -y {curl,wget,stow,build-essential,unzip,tree,dialog,bc}
+sudo apt-get install -qq -y curl wget stow build-essential unzip tree dialog bc fakeroot
```
wizard-28, line 146: `fakeroot` package hasn't been updated since 2014. Highly unlikely people will have an outdated version.

And `bc` is even older, it was last updated in 2011

misc/scripts/install-local.sh
```diff
@@ -311,7 +316,7 @@ if [[ -n $patch ]]; then
 fi
 
 export pkgdir="/usr/src/pacstall/$name"
-prepare
+fakeroot -- prepare
```
wizard-28, line 319: From the man page I think it's trying to run a "command" called `prepare` not a function.

See this syntax: ` fakeroot [-l|--lib library] [--faked faked-binary] [--] [command]`

Henryws: Actually I tested with a mini script before putting it into pacstall. And it worked, but it doesn't in pacstall. That's what I'm trying to figure out

Henryws: You have to run `export -f loca`

install.sh
D-Brox: We also don't use `dialog` since 1.5 release

wizard-28 (approved):

PR #153, opened 2021-07-23
Feat: use virtual packages for better apt integration & implement $gives

rain1598:
# Purpose
Unlike with the AUR, Pacstall does not provide the system's native package manager (in this case dpkg and apt) information about its packages. Information such as dependencies and conflicting packages which could potentially cause collisions between these package managers if not provided.
# Approach
This pull request adds the method makeVirtualDeb() in install-local.sh which creates an empty .deb package named `name`-pacstall with only the control file. The control file will then contain package and dependency info that could be read by apt. The control file will also contain `gives`, a variable which is included in .pacscripts but has never been implemented ("reverse deprecated").


wizard-28:
Umm... Can you read `CONTRIBUTING.md` file first... It has some guidelines...

---

Also follow the [pull request template](https://github.com/pacstall/pacstall/blob/develop/.github/pull_request_template.md)

Henryws:
Also can you fall back to `$name` if `$gives` is not there?

misc/scripts/install-local.sh
```diff
@@ -110,6 +110,32 @@ function log() {
...
 
+
+function makeVirtualDeb {
+	# creates empty .deb package (with only the control file) for apt integration
+	# implements $(gives) variable
+	# Note: I only put "development" in "Section" because I had to put something there
+	sudo mkdir -p $SRCDIR/$name-pacstall/DEBIAN
+	echo "Package: $name-pacstall
+Version: $version
+Depends: ${depends//' '/' | '}
```
D-Brox, line 121: Add optdepends to depends list if user decides to install them

D-Brox: Also, since apt would install the dependencies, we don't need to explicitly install them.

misc/scripts/install-local.sh
```diff
@@ -371,6 +397,8 @@ if [[ $? -ne 0	 ]]; then
 	return 1
 fi
 
+makeVirtualDeb
```
D-Brox, line 400: Move this before the dependencies install to keep track of unneeded packages after uninstall

misc/scripts/install-local.sh
D-Brox: Also, add a check on cleanup if it installation failed to remove the deb

misc/scripts/install-local.sh
D-Brox: And remove the apt installs of the dependencies, since the deb install would do that for us

D-Brox (changes requested):

misc/scripts/install-local.sh
rain1598: I don't think I should put optdepends in depends but I'll put them in suggests so uninstalling them is also optional

D-Brox: Ye, that works well for us

misc/scripts/install-local.sh
```diff
@@ -216,6 +251,8 @@ if [[ $NOBUILDDEP -eq 0 ]]; then
 	fi
 fi
 
+makeVirtualDeb
```
D-Brox, line 254: Add fancy messages for creating dummy package and for installing dependencies

D-Brox: Also, add the dummy deb removal in `trap_ctrl_c` if the dummy has already been installed. A simple variable check should do the job

D-Brox: And add the dummy deb removal before every `return 1` that comes after the `makeVirtualDeb` call

D-Brox (changes requested):

misc/scripts/remove.sh
```diff
@@ -85,6 +85,8 @@ case "$url" in
 	;;
 esac
 
+sudo dpkg -r "$name-pacstall" # removes virtual .deb package
```
D-Brox, line 88: This should be inside the `*)` case after, removing package, with a fancy message about removing the dummy

misc/scripts/install-local.sh
```diff
@@ -216,6 +255,8 @@ if [[ $NOBUILDDEP -eq 0 ]]; then
...
```
D-Brox, line 258: The function should only be called if the package is not a `-deb`
Use `if [[ "$url" == *".deb" ]]`

D-Brox (changes requested):
After this, we can merge

D-Brox (approved):

PR #155, opened 2021-07-24
Feature/better logging

Henryws:
# Purpose

Currently, the error logs are very sparse and should show more info about said error

# Approach

Using [this method](https://stackoverflow.com/a/3215760/13449010), you can output to stdout and to a file.

# Progress

- [x] Better logging for `build`
- [x] Better logging for `prepare`
- [x] Better logging for `install`

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -204,7 +204,7 @@ fi
 if [[ -n "$ppa" ]]; then
 	for i in "${ppa[@]}"; do
 		# Add ppa, but ppa bad I guess
-		sudo add-apt-repository ppa:"$i"
+		sudo add-apt-repository ppa:"$i" 2>&1 | tee -a "$LOGFILE"
 	done
 fi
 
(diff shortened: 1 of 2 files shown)


misc/scripts/install-local.sh
```diff
@@ -363,19 +373,19 @@ if ! command -v stow > /dev/null; then
 fi
 
 # Magic time. This installs the package to /, so `/usr/src/pacstall/foo/usr/bin/foo` -> `/usr/bin/foo`
-sudo stow --target="/" "$PACKAGE"
+sudo stow --target="/" "$PACKAGE" 2>&1 | tee -a "$LOGFILE"
 # stow will fail to symlink packages if files already exist on the system; this is just an error
 if [[ $? -ne 0	 ]]; then
```
wizard-28, line 378:
```suggestion
if [[ $? -ne 0 ]]; then
```

Remove unholy space

wizard-28 (changes requested):

PR #161, opened 2021-07-26
Fix: Caching

### LINKED ISSUE #160: Pacstall doesn't redownload scripts
**Describe the bug**
When downloading a pacscript, it won't overwrite the old script in `~/.cache/pacstall/` 

**To Reproduce**
Download a script, wait for an update to the script, and see that it won't overwrite the old file

**Expected behavior**
For it to redownload the script every time pacstall downloads a pacscript

**Desktop (please complete the following information):**

- OS + version: Linux Mint
- Pacstall Version + branch: develop

wizard-28:
# Purpose

Currently the caching in `download.sh` caches pacscripts to  `~/.cache/pacstall`, and it doesn't get cleaned after `install-local.sh` ends 

# Approach

This changes the caching directory to `/tmp/pacstall/pacscripts/`, and cleans it after `install-local` ends

# Progress

- [x] Change caching directory
- [x] Overwrite the file before downloading it
- [x] Change download dir based on `type` 

# Addendum

Fixes #160

### DIFF
--- a/misc/scripts/download.sh
+++ b/misc/scripts/download.sh
@@ -33,7 +33,6 @@ fi
 if curl --output /dev/null --silent --head --fail "$URL" ; then
 	if [[ "$type" = "install" ]]; then
 		mkdir -p "/tmp/pacstall/pacscripts/" && cd "/tmp/pacstall/pacscripts/"
-		mkdir -p "$PACKAGE" && cd "$PACKAGE"
 	fi
 
 	download "$URL" > /dev/null 2>&1
--- a/misc/scripts/download.sh
+++ b/misc/scripts/download.sh
@@ -31,8 +31,10 @@ if ! wget -q --tries=10 --timeout=20 --spider https://github.com; then
 fi
 
 if curl --output /dev/null --silent --head --fail "$URL" ; then
-	mkdir -p "/tmp/pacstall/pacscripts/" && cd "/tmp/pacstall/pacscripts/"
-	mkdir -p "$PACKAGE" && cd "$PACKAGE"
+	if [[ "$type" = "install" ]]; then
+		mkdir -p "/tmp/pacstall/pacscripts/" && cd "/tmp/pacstall/pacscripts/"
+		mkdir -p "$PACKAGE" && cd "$PACKAGE"
+	fi
 
 	download "$URL" > /dev/null 2>&1
 	return 0
--- a/misc/scripts/download.sh
+++ b/misc/scripts/download.sh
@@ -31,7 +31,7 @@ if ! wget -q --tries=10 --timeout=20 --spider https://github.com; then
 fi
 
 if curl --output /dev/null --silent --head --fail "$URL" ; then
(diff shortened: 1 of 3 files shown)


Henryws (approved):

D-Brox:
Also nuke `.cache/pacstall` if it exists in `update.sh`

misc/scripts/install-local.sh
```diff
@@ -28,6 +28,8 @@ function cleanup () {
 	if [ -f /tmp/pacstall-optdepends ]; then
 		sudo rm /tmp/pacstall-optdepends
 	fi
+
+	rm -rf "/tmp/pacstall/pacscripts/"
```
D-Brox, line 32: We already nuke `/tmp/pacstall`, there is no need to nuke it again. It'll just print an error

misc/scripts/download.sh
```diff
@@ -31,8 +31,10 @@ if ! wget -q --tries=10 --timeout=20 --spider https://github.com; then
 fi
 
 if curl --output /dev/null --silent --head --fail "$URL" ; then
-	mkdir -p "$HOME/.cache/pacstall" && cd "$HOME"/.cache/pacstall/
-	mkdir -p "$PACKAGE" && cd "$PACKAGE"
+	if [[ "$type" = "install" ]]; then
+		mkdir -p "/tmp/pacstall/pacscripts/" && cd "/tmp/pacstall/pacscripts/"
+		mkdir -p "$PACKAGE" && cd "$PACKAGE"
```
D-Brox, line 36: I don't think we need to use `mkdir -p "$PACKAGE" && cd "$PACKAGE"` anymore in install

misc/scripts/download.sh
```diff
@@ -31,8 +31,10 @@ if ! wget -q --tries=10 --timeout=20 --spider https://github.com; then
...
+	fi
```
D-Brox, line 37: Needs an `else` for when `$type == "download` to specify where to download the pacscript

D-Brox (changes requested):

misc/scripts/install-local.sh
wizard-28: alright, on it.

misc/scripts/download.sh
wizard-28: why?

misc/scripts/download.sh
wizard-28: @Henryws says when `-D` flag is active, then just download it in `pwd`, so not specifying an `else` will automatically download it in `pwd`

misc/scripts/download.sh
D-Brox: Because of how we source the pacscript, it doesn't matter where it is, only that we are in the right dir

wizard-28: Alright, you know the best! On it.

misc/scripts/download.sh
D-Brox: Ok

D-Brox (approved):

NarutoXY (approved):

PR #162, opened 2021-07-26
Feature/better logging

Henryws:
# Purpose

Right now, the logging feature is very underpowered, as it only shows the error, so it's not really a log, just text. This will log important output to the log also

# Approach

With [this method](https://stackoverflow.com/a/3215760/13449010), you can output to the terminal and a file

# Progress

- [x] All important functions in `install-local.sh`
- [x] `remove.sh`

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -238,7 +238,7 @@ function hashcheck() {
 
 fancy_message info "Installing dependencies"
 sudo apt-get install -y -qq -o=Dpkg::Use-Pty=0 $depends 2>&1 | tee -a "$LOGFILE"
-if [[ $? -ne 0	 ]]; then
+if [[ $? -ne 0 ]]; then
 	fancy_message error "Failed to install dependencies"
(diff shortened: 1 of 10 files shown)


misc/scripts/install-local.sh
```diff
@@ -236,7 +237,7 @@ function hashcheck() {
 }
 
 fancy_message info "Installing dependencies"
-sudo apt-get install -y -qq -o=Dpkg::Use-Pty=0 $depends
+sudo apt-get install -y -qq -o=Dpkg::Use-Pty=0 $depends 2>&1 | tee -a "$LOGFILE"
 if [[ $? -ne 0	 ]]; then
```
wizard-28, line 241:
```suggestion
if [[ $? -ne 0 ]]; then
```

Remove unholy tab

wizard-28 (changes requested):

D-Brox:
Since it's not just an error log now, I'd suggest changing the dir name

Issue #168, opened 2021-07-31
Add option to purge


Korkmatik:
**Is your feature request related to a problem? Please describe.**
Currently when uninstalling with `pacstall -R <package>` it will perform an `apt remove <package-gives>` if it is a deb package.

**Describe the solution you'd like**
 I think we could also add an option that will run `apt purge <package-gives>`. For example by running `pacstall -Rp <package>` or `pacstall -P <package>` or `pacstall --purge <package>`

**Describe alternatives you've considered**
Running `apt purge` manually

Henryws:
Just so you know, the `-P` option is used for unattended installs (example: `pacstall -P -I neofetch`)

Korkmatik:
> Just so you know, the `-P` option is used for unattended installs (example: `pacstall -P -I neofetch`)

Ok thank you!

wizard-28:
How about `-R` purging by default?

D-Brox:
It's possible now that we implemented the `apt` integration in 1.6

Henryws:
We did this a while back.

D-Brox:
#505

PR #169, opened 2021-07-31
Add: Purge

NarutoXY:
# Purpose
Fixes https://github.com/pacstall/pacstall/issues/168

# Approach

Make `purge.sh`

# Progress

- [x] Purge

# Addendum

https://github.com/pacstall/pacstall/issues/168

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -347,7 +347,7 @@ if [[ -n $optdepends ]]; then
 	if [[ $answer -eq 1 ]]; then
 		for items in "${optdepends[*]}"; do
 			# output the name of the apt thing without the description, EI: `foo: not bar` -> `foo`
-			printf "%s\n" "${optdepends[@]}" | cut -f1 -d":" | tr '\n' ' ' >> /tmp/pacstall-optdepends
+			printf "%s\n" "${optdepends[@]}" | awk -F': ' '{print $1}' | tr '\n' ' ' >> /tmp/pacstall-optdepends
 			# Install
 			sudo apt-get install -y -qq $(cat /tmp/pacstall-optdepends)
 		done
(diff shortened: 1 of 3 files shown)


misc/scripts/purge.sh
```diff
@@ -0,0 +1,64 @@
...
+		if ! sudo apt purge "$gives" 2>/dev/null; then
+			fancy_message warn "Failed to remove the package"
+			error_log 1 "remove $PACKAGE"
+			return 1
+		fi
+		return 0
+	;;
+
+	*)
+		fancy_message error "Currently purge functionality is only available for deb packages."
```
D-Brox, line 57: If we use the dummy deb package we can purge the dependencies tho

pacstall
```diff
@@ -211,6 +211,7 @@ while [[ ! "$1" == "--" ]]; do
 			-Il --install-local, Installs a local package
 			-S --search, Search for package
 			-R --remove, Removes package
+			-Rp --remove-purge, Purges package (currently only deb packages are supported)
```
D-Brox, line 214: `--purge` makes more sense

misc/scripts/install-local.sh
```diff
@@ -347,7 +347,7 @@ if [[ -n $optdepends ]]; then
 	if [[ $answer -eq 1 ]]; then
 		for items in "${optdepends[*]}"; do
 			# output the name of the apt thing without the description, EI: `foo: not bar` -> `foo`
-			printf "%s\n" "${optdepends[@]}" | cut -f1 -d":" | tr '\n' ' ' >> /tmp/pacstall-optdepends
+			printf "%s\n" "${optdepends[@]}" | awk -F': ' '{print $1}' | tr '\n' ' ' >> /tmp/pacstall-optdepends
```
D-Brox, line 350: This is being deprecated with the dummy deb apt integration, no need to change it

PR #172, opened 2021-08-02
fix (workflow): rename brave package

Henryws:
# Purpose

Change the brave deb name


NarutoXY:
Test succeed 👍🏻 so merging

PR #179, opened 2021-08-04
Feat: add postrm for apt integration

D-Brox:
# Purpose

Pacstall dummy packages currently cannot be removed with apt

# Approach

Adds a postrm file that calls pacstall


misc/scripts/install-local.sh
```diff
@@ -119,7 +119,7 @@ function makeVirtualDeb {
 	# implements $(gives) variable
 	fancy_message info "Creating dummy package"
 	sudo mkdir -p "$SRCDIR/$name-pacstall/DEBIAN"
-	printf "Package: $name-pacstall\n" | sudo tee "$SRCDIR/$name-pacstall/DEBIAN/control" > /dev/null
+	printf "Package: $name\n" | sudo tee "$SRCDIR/$name-pacstall/DEBIAN/control" > /dev/null
 	if [[ $version =~ "^[0-9]" ]]; then
```
D-Brox, line 123: Should be `if [[ $version =~ ^[0-9] ]]; then`

PR #183, opened 2021-08-09
Feat: Initial `install.sh` revamp

wizard-28:
# Purpose

The current `install.sh` isn't the greatest installation script. There are many Code Quality issues and bugs in it.

# Approach

This PR is supposed to fix all those issues.

# Progress

- [x] `chown` `/tmp/pacstall` directory
- [x] Refactor `$LOGNAME` -> `$(logname)`
- [x] Fix `ask` function
- [x] Fix capitalization / grammer
- [x] Improve Code Quality
- [x] Remove useless code

# Addendum

This is the initial revamp of the script, I guess more PRs will be made later with more revamps.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -39,7 +39,7 @@ function fancy_message() {
 	# 0 - info
 	# 1 - warning
 	# 2 - error
-	if [ -z "${1}" ] || [ -z "${2}" ]; then
+	if [[ -z "${1}" ]] || [[ -z "${2}" ]]; then
 		return
 	fi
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -62,134 +66,114 @@ function ask() {
...
-if command -v fish &>/dev/null; then
-	sudo wget -q --show-progress --progress=bar:force -O /usr/share/fish/vendor_completions.d/pacstall.fish https://raw.githubusercontent.com/pacstall/pacstall/master/misc/completion/fish &
-fi
+wget -q --show-progress --progress=bar:force -O "/usr/share/bash-completion/completions/pacstall" "https://raw.githubusercontent.com/pacstall/pacstall/master/misc/completion/bash" &
+wget -q --show-progress --progress=bar:force -O "/usr/share/fish/vendor_completions.d/pacstall.fish" "https://raw.githubusercontent.com/pacstall/pacstall/master/misc/completion/fish" &
 
 wait
 
-sudo chmod +x /bin/pacstall
+chmod +x "/bin/pacstall"
```
Henryws, line 177: This will fail in a non sudo environment

Henryws (changes requested):

install.sh
wizard-28: The script is running using `sudo` everytime. See the installation snippet
```bash
sudo bash -c "$(curl -fsSL https://git.io/JsADh || wget -q https://git.io/JsADh -O -)"
```

Henryws: > The script is running using `sudo` everytime. See the installation snippet
> 
> ```shell
> sudo bash -c "$(curl -fsSL https://git.io/JsADh || wget -q https://git.io/JsADh -O -)"
> ```

small 🧠 moment

Henryws (approved):

install.sh
```diff
@@ -62,134 +66,114 @@ function ask() {
...
-i=1
-sp="/-\|"
-printf ' '
-while [ -d /proc/$PID ]
-do
-	sleep 0.1
-	printf "\b${sp:i++%${#sp}:1}"
-done
-if [[ $? -eq 1 ]] ; then
+if ! nm-online -qx; then
```
NarutoXY, line 117: What if someone don't has `nm-online`? Some use `systemd-networkd` and some also use `connman`

NarutoXY (changes requested):

install.sh
wizard-28: `NetworkManager` (which is the main app) ships with every linux distribution. It is independent of `systemd-networkd` ( They are totally different apps, and can work when either of them is not present in the system ). So it should work in every distribution.

NarutoXY: Well `Clear Linux` and many more use `systemd-networkd` and don't ship with `networkmanager`. And some people may remove `networkmanager` from their system and use `systemd-networkd` so you can set `wget` one as fallback

NarutoXY (approved):

PR #191, opened 2021-08-15
Feat: Added argument splitter for grouped short options

hwittenborn:
# Purpose

Previously, arguments for short options passed to Pacstall were not allowed to be grouped together (i.e. you had to do `-P -I` instead of `-PI`, as `-PI` wouldn't be recognized as an option).

# Approach

Now, any options that start with a single hyphen are automatically split on every occurrence of a capital letter.

Likewise:

`-PI` would become `-P -I`

`-PIl` would become `-P -Il`

`-IlP` would become `-Il -P`


hwittenborn:
Looks like I forgot to remove the echo and exit commands :P

I'll get that fixed later today

pacstall
```diff
@@ -205,6 +205,26 @@ fi
...
+		continue
+	fi
+
+	# Add all arguments to the list of arguments.
+	# We remove the '-' prefix as we'll add it later.
+	for j in $(echo "${i}" | sed 's|^-||' | sed 's|[[:upper:]]| &|g'); do
+		argument_list+=("-${j}")
+	done
+done
+
```
D-Brox, line 224: If the new arguments list has a `-P` you should move it to the start so that the prompts are disabled properly

D-Brox (changes requested):

D-Brox:
Also, can you fetch upstream so that the update script is fixed on your branch? That way we can test it more easily

pacstall
hwittenborn: Done

hwittenborn:
> Also, can you fetch upstream so that the update script is fixed on your branch? That way we can test it more easily

All done

D-Brox (approved):

PR #192, opened 2021-08-15
Feat: Unset variables in `install-local.sh` cleanup

echometerain:
# Purpose

When installing multiple programs at once, variables left over may affect later installs

# Approach

It unsets the affected variables in cleanup

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -29,8 +29,7 @@ function cleanup () {
 	if [ -f /tmp/pacstall-optdepends ]; then
 		sudo rm /tmp/pacstall-optdepends
 	fi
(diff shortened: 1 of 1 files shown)


Henryws:
I think the reason why it's failing is because of `set -e`. What you should do is a `for i` loop that checks if one of the possible variables exists, then unset them

misc/scripts/install-local.sh
```diff
@@ -29,6 +29,24 @@ function cleanup () {
...
+	unset hash
+	unset removescript
+	unset optdepends
+	unset ppa
+	unset maintainer
+	unset pacdeps
+	unset patch
+	unset PACPATCH
+	unset NOBUILDDEP
+	unset optinstall
```
D-Brox, line 49: You can unset all in the same line

D-Brox (changes requested):

D-Brox (approved):

PR #197, opened 2021-08-17
Replace $(logname) with $LOGNAME

echometerain:
# Purpose

A lot of systems have a broken logname program

# Approach

I replaced $(logname) with $LOGNAME but the code replaces $LOGNAME with $(logname) unless $(logname) fails

# Addendum

That was a terrible way to phrase it


Henryws:
Did you test this on any packages?

echometerain:
yes, lyrebird and exploitdb

PR #199, opened 2021-08-18
Replace tee with $control for better readability

echometerain:
# Purpose

` | sudo tee -a "$SRCDIR/$name-pacstall/DEBIAN/control" > /dev/null` shouldn't be called every other line due to speed and readability.

# Approach

Uses $control as a temporary control file and writes to `$SRCDIR/$name-pacstall/DEBIAN/control` once


Henryws:
What packages did you test this on?

echometerain:
I give up

PR #202, opened 2021-08-20
Feat: handle tarball programs

echometerain:
# Purpose
A lot of programs ship as tarballs

# Approach

Unzip the tar like the .zip

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -397,9 +397,7 @@ case "$url" in
 		download "$url"
 		# I think you get it by now
 		hashcheck "${url##*/}"
-		sudo tar -xf "${url##*/}" 1>&1 2>/dev/null
-		cd ./*/ 2>/dev/null
-		export srcdir="/tmp/pacstall/$PWD"
+		export srcdir="/tmp/pacstall"
 		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
 	;;
 esac
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -358,6 +358,19 @@ case "$url" in
 		# Make the directory available for users
 		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
 	;;
+	*.tar*)
+		download "$url"
+		# hash the file
+		hashcheck "${url##*/}"
+		# unzip file
+		sudo tar -xf "${url##*/}" 1>&1 2>/dev/null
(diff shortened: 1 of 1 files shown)


misc/scripts/install-local.sh
```diff
@@ -358,6 +358,19 @@ case "$url" in
...
+		hashcheck "${url##*/}"
+		# unzip file
+		sudo tar -xf "${url##*/}" 1>&1 2>/dev/null
+		# cd into it
+		cd ./*/
+		# export srcdir
+		export srcdir="/tmp/pacstall/$PWD"
+		# Make the directory available for users
+		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
+	;;
```
D-Brox, line 373: Technically this is already done in `*)`, but I prefer it this way. Remove the `tar` and the `cd` from `*)` and we could merge.

D-Brox (changes requested):

misc/scripts/install-local.sh
```diff
@@ -393,6 +393,19 @@ case "$url" in
 		# Make the directory available for users
 		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
 	;;
+	*.tar*)
+		download "$url"
+		# hash the file
+		hashcheck "${url##*/}"
```
wizard-28, line 399:
```suggestion	
		if ! hashcheck "${url##*/}"; then
			return 1
		fi
```

misc/scripts/install-local.sh
```diff
@@ -425,12 +438,7 @@ case "$url" in
 		download "$url"
 		# I think you get it by now
 		hashcheck "${url##*/}"
```
wizard-28, line 440:
```suggestion	
		if ! hashcheck "${url##*/}"; then
			return 1
		fi
```

misc/scripts/install-local.sh
wizard-28: Why are we adding a separate case for `.tar`? In this method in future we may have to add different cases for `.gz`, `.xz` and other types of `tar`s (which were currently automatically handled by `*)`)

wizard-28 (changes requested):

NarutoXY:
Well @wizard-28 what are plans for this pr?

wizard-28:
I don't really know... Like I said. We are already unzipping tars...

NarutoXY:
Yea this is like a redundant pr. We already are unzipping tars. Closing for now

PR #208, opened 2021-08-21
Fix(scripts): add missing cleanups

D-Brox:
# Purpose

There where missing cleanups and the hashcheck didn't error out

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -34,7 +34,7 @@ function cleanup () {
 function trap_ctrlc () {
 	echo ""
 	fancy_message warn "Interupted, cleaning up"
-	if dpkg-query -W -f='${Status}' "$name" 2> /dev/null | grep -q "ok installed" ; then
+	if dpkg-query -W -f='${Status}' "$name" 2> /dev/null | grep -q -E "ok installed|ok unpacked" ; then
 		sudo dpkg -r --force-all "$name" > /dev/null
 	fi
 	cleanup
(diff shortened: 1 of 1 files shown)


misc/scripts/install-local.sh
```diff
@@ -361,6 +381,9 @@ case "$url" in
 		download "$url"
 		# hash the file
 		hashcheck "${url##*/}"
+		if [[ $? -ne 0 ]]; then
+			return 1
+		fi
```
wizard-28, line 386:
```suggestion	
		if ! hashcheck "${url##*/}"; then
			return 1
		fi
```

misc/scripts/install-local.sh
```diff
@@ -373,6 +396,9 @@ case "$url" in
 	*.deb)
 		download "$url"
 		hashcheck "${url##*/}"
+		if [[ $? -ne 0 ]]; then
+			return 1
+		fi
```
wizard-28, line 401:
```suggestion	
		if ! hashcheck "${url##*/}"; then
			return 1
		fi
```

misc/scripts/install-local.sh
```diff
@@ -382,20 +408,26 @@ case "$url" in
...
 			return 1
 		fi
 	;;
 	*)
 		download "$url"
 		# I think you get it by now
 		hashcheck "${url##*/}"
+		if [[ $? -ne 0 ]]; then
+			return 1
+		fi
```
wizard-28, line 430:
```suggestion	
		if ! hashcheck "${url##*/}"; then
			return 1
		fi
```

misc/scripts/install-local.sh
Henryws: yeah, this would be nice, because it looks nicer, and it gives minor speed improvements (even if it's very small

PR #214, opened 2021-08-22
rem(install-local): View question

wizard-28:
# Purpose

Currently we are asking the user to view, then to edit the pacscript. This can be done in a one step process instead of a two step one.

# Approach

Remove viewing question. Directly ask to view/edit the pacscript

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -205,13 +205,7 @@ fi' | sudo tee "$SRCDIR/$name-pacstall/DEBIAN/postrm" >"/dev/null"
 	return 0
 }
 
-
-ask "Do you want to view the pacscript first" N
-if [[ $answer -eq 1 ]]; then
-	less "$PACKAGE".pacscript
-fi
-
-ask "Do you want to edit the pacscript" N
+ask "Do you want to view/edit the pacscript" N
 if [[ $answer -eq 1 ]]; then
 	if [[ -n $PACSTALL_EDITOR ]]; then
 		$PACSTALL_EDITOR "$PACKAGE".pacscript


misc/scripts/install-local.sh
```diff
@@ -205,13 +205,7 @@ fi' | sudo tee "$SRCDIR/$name-pacstall/DEBIAN/postrm" >"/dev/null"
...
 }
 
-
-ask "Do you want to view the pacscript first" N
-if [[ $answer -eq 1 ]]; then
-	less "$PACKAGE".pacscript
-fi
-
-ask "Do you want to edit the pacscript" N
+ask "Do you want to view/edit the pacscript" N
```
NarutoXY, line 208: After editing, you may ask user whether to decline the installation or not. 
```bash
ask "Do you want to install the package?" Y
if [[ $answer -eq 1 ]]; then
	return 1
fi
```

NarutoXY (changes requested):

misc/scripts/install-local.sh
wizard-28: They can <kbd>ctrl</kbd> + <kbd>C</kbd>

NarutoXY: Well they can but `pacstall` should ask this as after all it is just user repo and it's a good practice

wizard-28: Hmm, so think the other way, if the user actually wants to install the software after editing the script, they now have to answer 2 questions.

NarutoXY: Just press enter and done

wizard-28: Yeah they can also just press <kbd>ctrl</kbd> + <kbd>C</kbd> and be done, it's more ergonomic as most of the time the users would want to install the software, so in that case they have to press 1 extra key every time, as opposed to when they don't want to install the software (rare)

NarutoXY: Well if I quit the editor then the installation begins without any consent so...

wizard-28: Hmm... I think running `pacstall -I <script>` is consent enough

Henryws (approved):

NarutoXY (approved):

PR #235, opened 2021-09-02
feat(main): `chown` `STGDIR`

Henryws:
# Purpose

Fixes https://github.com/pacstall/pacstall-programs/issues/221

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -440,7 +440,7 @@ if [[ -n $patch ]]; then
 fi
 
 export pkgdir="/usr/src/pacstall/$name"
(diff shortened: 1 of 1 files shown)


wizard-28:
Well it doesn't fix https://github.com/pacstall/pacstall-programs/issues/221. As it was a problem with `SRCDIR` not `STGDIR`, but since we were aiming to `chown` `STGDIR` sometime in the future anyway. I'll accept this PR.

wizard-28:
Secondly, I would prefer `chown`ing stuff in the installer, but since we are having a python3 rewrite I'll solve this problem there in the installer.

PR #246, opened 2021-09-11
Feature/back to sudo

Henryws:
## Purpose

Trying to make pacstall user installable has been a failure, so lets go back to using sudo

## Approach

Adds sudo back to where it should be

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -366,8 +366,6 @@ case "$url" in
 		cd ./*/ 2> /dev/null
 		# The srcdir is /tmp/pacstall/foo
 		readonly srcdir="/tmp/pacstall/$PWD" ; export srcdir
-		# Make the directory available for users
-		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
 		# Check the integrity
 		git fsck --full
 	;;
@@ -384,8 +382,6 @@ case "$url" in
 		cd ./*/ 2> /dev/null
 		# export srcdir
 		readonly srcdir="/tmp/pacstall/$PWD" ; export srcdir
-		# Make the directory available for users
-		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
 	;;
 	*.deb)
 		download "$url"
@@ -425,23 +421,20 @@ case "$url" in
 		sudo tar -xf "${url##*/}" 1>&1 2>/dev/null
 		cd ./*/ 2>/dev/null
 		readonly srcdir="/tmp/pacstall/$PWD" ; export srcdir
-		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
 	;;
 esac
 
 if [[ -n $patch ]]; then
 		fancy_message info "Downloading patches"
-		mkdir -p PACSTALL_patchesdir
+		sudo mkdir -p PACSTALL_patchesdir
 	for i in "${patch[@]}"; do
-		wget -q "$i" -P PACSTALL_patchesdir &
+		sudo wget -q "$i" -P PACSTALL_patchesdir &
 	done
 	wait
 	readonly PACPATCH=$(pwd)/PACSTALL_patchesdir ; export PACKAGE
 fi
 
 readonly pkgdir="/usr/src/pacstall/$name" ; export pkgdir
-sudo chown -R "$USER":"$USER" /usr/src/pacstall
-
 # fakeroot is weird but this method works
 # create tmp variable that is the output of what prepare function is (it prints out function)
 if ! command -v fakeroot > /dev/null; then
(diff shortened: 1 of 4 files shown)


NarutoXY (changes requested):
We can fix those issues. Remove sudo.

misc/scripts/install-local.sh
```diff
@@ -208,11 +208,11 @@ fi' | sudo tee "$SRCDIR/$name-pacstall/DEBIAN/postrm" >"/dev/null"
 ask "Do you want to view/edit the pacscript" N
 if [[ $answer -eq 1 ]]; then
 	if [[ -n $PACSTALL_EDITOR ]]; then
-		$PACSTALL_EDITOR "$PACKAGE".pacscript
+		sudo $PACSTALL_EDITOR "$PACKAGE".pacscript
```
NarutoXY, line 211: we can use `sudoedit` instead of all these if statements. `sudo $EDITOR` is vague and dangerous as it provides editor with root access. but `sudoedit` edits a temp file using $EDITOR to prevent misuse of privileges by the editor

NarutoXY (changes requested):

NarutoXY (approved):

misc/scripts/install-local.sh
wizard-28: Why do we even need `sudo` to edit these files? The pacscripts are downloaded in `/tmp/pacstall` which is owned by the owner.

misc/scripts/install-local.sh
```diff
@@ -208,11 +208,11 @@ fi' | sudo tee "$SRCDIR/$name-pacstall/DEBIAN/postrm" >"/dev/null"
...
 	elif [[ -n $EDITOR ]]; then
-		$EDITOR "$PACKAGE".pacscript
+		sudo $EDITOR "$PACKAGE".pacscript
```
wizard-28, line 213: Why do we even need `sudo` to edit these files? The pacscripts are downloaded in `/tmp/pacstall` which is owned by the owner.

misc/scripts/install-local.sh
```diff
@@ -208,11 +208,11 @@ fi' | sudo tee "$SRCDIR/$name-pacstall/DEBIAN/postrm" >"/dev/null"
...
 	elif [[ -n $VISUAL ]]; then
-		$VISUAL "$PACKAGE".pacscript
+		sudo $VISUAL "$PACKAGE".pacscript
```
wizard-28, line 215: Why do we even need `sudo` to edit these files? The pacscripts are downloaded in `/tmp/pacstall` which is owned by the owner.

misc/scripts/install-local.sh
```diff
@@ -366,8 +366,6 @@ case "$url" in
 		cd ./*/ 2> /dev/null
 		# The srcdir is /tmp/pacstall/foo
 		readonly srcdir="/tmp/pacstall/$PWD" ; export srcdir
-		# Make the directory available for users
-		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
```
wizard-28, line 370: Why remove this?

misc/scripts/install-local.sh
```diff
@@ -384,8 +382,6 @@ case "$url" in
 		cd ./*/ 2> /dev/null
 		# export srcdir
 		readonly srcdir="/tmp/pacstall/$PWD" ; export srcdir
-		# Make the directory available for users
-		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
```
wizard-28, line 388: Why remove this?

misc/scripts/install-local.sh
```diff
@@ -425,23 +421,20 @@ case "$url" in
 		sudo tar -xf "${url##*/}" 1>&1 2>/dev/null
 		cd ./*/ 2>/dev/null
 		readonly srcdir="/tmp/pacstall/$PWD" ; export srcdir
-		sudo chown -R "$LOGNAME":"$LOGNAME" . 2>/dev/null
```
wizard-28, line 428: Why remove this?

misc/scripts/install-local.sh
```diff
@@ -425,23 +421,20 @@ case "$url" in
...
 	;;
 esac
 
 if [[ -n $patch ]]; then
 		fancy_message info "Downloading patches"
-		mkdir -p PACSTALL_patchesdir
+		sudo mkdir -p PACSTALL_patchesdir
```
wizard-28, line 429: Same, doesn't need `sudo`, as it's owned by user.

misc/scripts/install-local.sh
```diff
@@ -425,23 +421,20 @@ case "$url" in
...
 	for i in "${patch[@]}"; do
-		wget -q "$i" -P PACSTALL_patchesdir &
+		sudo wget -q "$i" -P PACSTALL_patchesdir &
```
wizard-28, line 431: Same, doesn't need `sudo`, as it's owned by user.

wizard-28 (changes requested):

wizard-28:
The LoC which I've reviewed saying things along the lines of "Why need `sudo` here?". Those lines didn't have `sudo` in 1.5.1 & previous versions. `/tmp/pacstall` has always been owned by the user... In `1.6` we were experimenting with `chown`ing `/usr/src/pacstall`, I think you've mixed up those two directories.

PR #247, opened 2021-09-12
Prune Attempt

NarutoXY:
## Purpose

repo is too big

## Approach

prune

## Progress

- [x] prune


NarutoXY:
rebase on this instead of merging

wizard-28:
Hey so, after pruning all the previous commits which were `verified` are now `unverified`?

wizard-28:
> ## Purpose
> 
> repo is too big
> ## Approach
> 
> prune
> ## Progress
> 
>     * [x]  prune

Why do we need to prune it, we can shallow clone it...

PR #264, opened 2021-09-21
feat(scripts): Add more checks to the check function

Henryws:
## Purpose

Currently the `check` function doesn't do much, so I added more checks that look for vital variables and alert the user if there is an issue

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -44,13 +44,6 @@ function trap_ctrlc () {
 
 # run checks to verify script works
 function checks() {
-	# curl url to check it exists
-	if curl --output /dev/null --silent --head --fail "$url" > /dev/null; then
-		fancy_message info "URL exists"
-	else
-		fancy_message error "URL doesn't exist"
-		return 1
-	fi
 
 	if [[ -z "$name" ]]; then
 		fancy_message error "Package does not contain name"
@@ -76,6 +69,13 @@ function checks() {
 	if [[ -z "$maintainer" ]]; then
 		fancy_message warn "Package does not have a maintainer"
 		fancy_message warn "It maybe no longer maintained. Please be advised."
+	fi
+	# curl url to check it exists
+	if curl --output /dev/null --silent --head --fail "$url" > /dev/null; then
+		fancy_message info "URL exists"
+	else
+		fancy_message error "URL doesn't exist"
+		return 1
 	fi
 }
 
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -52,6 +52,10 @@ function checks() {
 		return 1
 	fi
 
+	if [[ -z "$name" ]]; then
+		fancy_message error "Package does not contain name"
+		exit 1
+	fi
 	if [[ -z "$hash" ]]; then
 		# if the name doesn't end in -git, we put up the prompt
 		if ! [[ "$name" == "*-git" ]]; then
@@ -63,10 +67,6 @@ function checks() {
 	fi
 	if [[ -z "$version" ]]; then
 		fancy_message error "Package does not contain version"
-		exit 1
-	fi
-	if [[ -z "$name" ]]; then
-		fancy_message error "Package does not contain name"
 		exit 1
 	fi
 	if [[ -z "$url" ]]; then
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -53,7 +53,10 @@ function checks() {
 	fi
 
 	if [[ -z "$hash" ]]; then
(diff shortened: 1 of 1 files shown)


misc/scripts/install-local.sh
```diff
@@ -55,6 +55,21 @@ function checks() {
 	if [[ -z "$hash" ]]; then
 		fancy_message warn "Package does not contain a hash"
 	fi
```
wizard-28, line 57: Could we remove "Package does not contain a hash" message for `-git` pacscripts?

misc/scripts/install-local.sh
```diff
@@ -55,6 +55,21 @@ function checks() {
...
+		exit 1
+	fi
+	if [[ -z "$name" ]]; then
+		fancy_message error "Package does not contain name"
+		exit 1
+	fi
+	if [[ -z "$url" ]]; then
+		fancy_message error "Package does not contain URL"
+		exit 1
+	fi
```
wizard-28, line 72:
```suggestion
	fi
	if [[ -z "$maintainer" ]]; then
		fancy_message warn "Package does not have a maintainer"
		fancy_message warn "It maybe no longer maintained. Please be advised."
	fi
```

I think we could add a warning for missing `maintainer` fields. So if some non-`-git` pacscript gets no longer updated for quite some time. We could remove that field, and the users will be warned accordingly.

Or the maintainer themselves could remove the field if they no longer wish to maintain the script.

wizard-28 (changes requested):

misc/scripts/install-local.sh
Henryws: This could be one line

wizard-28: What do you suggest for the one liner?

wizard-28: Too much is cramped into a single line. I prefer 2 line approach.

misc/scripts/install-local.sh
```diff
@@ -52,8 +52,30 @@ function checks() {
...
 	fi
 
+	if [[ -z "$name" ]]; then
+		fancy_message error "Package does not contain name"
+		exit 1
+	fi
 	if [[ -z "$hash" ]]; then
-		fancy_message warn "Package does not contain a hash"
+		# if the name doesn't end in -git, we put up the prompt
+		if ! [[ "$name" == "*-git" ]]; then
```
wizard-28, line 61:
```suggestion
		if [[ "$name" != "*-git" ]]; then
```

Simplify condition.

wizard-28 (changes requested):

misc/scripts/install-local.sh
```diff
@@ -44,17 +44,36 @@ function trap_ctrlc () {
...
 function checks() {
+
+	if [[ -z "$name" ]]; then
+		fancy_message error "Package does not contain name"
+		exit 1
+	fi
+	if [[ -z "$hash" && "$name" != "*-git" ]]; then
+		fancy_message warn "Package does not contain a hash"
+	fi
+	if [[ -z "$description" ]]; then
```
NarutoXY, line 55: `description` is not necessary tbh

misc/scripts/install-local.sh
```diff
@@ -44,17 +44,36 @@ function trap_ctrlc () {
...
+	fi
+	if [[ -z "$version" ]]; then
+		fancy_message error "Package does not contain version"
+		exit 1
+	fi
+	if [[ -z "$url" ]]; then
+		fancy_message error "Package does not contain URL"
+		exit 1
+	fi
+	if [[ -z "$maintainer" ]]; then
```
NarutoXY, line 66: `maintainer` is also not the essential ones. We can show them as a warning

misc/scripts/install-local.sh
NarutoXY: oh wait did not notice you were not exiting. `maintainer` can be resolved but `description` should not show any warning as description is just for package maintainer's sanity

NarutoXY (commented):
All good. There is a suggestion tho

misc/scripts/install-local.sh
```diff
@@ -44,17 +44,36 @@ function trap_ctrlc () {
 
 # run checks to verify script works
 function checks() {
+
+	if [[ -z "$name" ]]; then
+		fancy_message error "Package does not contain name"
+		exit 1
+	fi
+	if [[ -z "$hash" && "$name" != "*-git" ]]; then
```
wizard-28, line 52:
```suggestion
	if [[ -z "$hash" && "$name" != *-git ]]; then
```

Fix `*-git` glob

wizard-28 (changes requested):

wizard-28 (approved):
Works in my short test.

PR #268, opened 2021-10-01
feat(scripts): Ability to specify target repo when installing

### LINKED ISSUE #126: Specify from which repo to install for use in scripting
Add the possibility to specify from which repo to install before searching then with `pacstall -I  repo-alias/package`

D-Brox:
## Purpose

Specify from which repo to install for use in scripting

## Approach

Add the possibility to specify from which repo to install before searching then with `pacstall -I package@user/reponame` for github and gitlab repos or `pacstall -I package@repourl` for other git-like repos.

## Addendum

Fixes #126

### DIFF
--- a/misc/scripts/search.sh
+++ b/misc/scripts/search.sh
@@ -54,7 +54,8 @@ if [[ $PACKAGE == *@* ]]; then
 		if [[ $URLNAME == $REPONAME ]]; then
 			PACKAGELIST=($(curl -s "$URL"/packagelist))
 			IDXSEARCH=$(printf "%s\n" "${PACKAGELIST[@]}" | grep -n "^${PACKAGE}$")
-			LEN=${#IDXSEARCH[@]}
+			LEN=($IDXSEARCH)
+			LEN=${#LEN[@]}
 			if [[ "$LEN" -eq 0 ]]; then
 				fancy_message warn "There is no package with the name $IRed${PACKAGE%%@*}$CYAN @ $REPONAME$NC"
 				error_log 3 "search $PACKAGE@$REPONAME"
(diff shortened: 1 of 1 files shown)


Henryws (approved):

wizard-28 (changes requested):
I think we should log the place from where the package was installed in `$LOGDIR/metadata/pkg`. Also print it using the `-Qi` command.

---

Also display 
```
[*] INFO: Downloading <colored>pkg</colored> from <colored>`url of repo`</colored>
```
when downloading from a non-official repo

---

img

Ask for the other repos too. Don't call it `repo pacstall`, call it `official repo`.

---

img

Don't make it a `ask` just leave it as a `info`

misc/scripts/search.sh
```diff
@@ -31,6 +31,47 @@ if [[ -n "$UPGRADE" ]]; then
...
+				error_log 3 "search $PACKAGE@$REPONAME"
+				return 1	
+			fi
+			export PACKAGE
+			export REPO=$URL
+			return 0
+		fi
+	done < "$STGDIR/repo/pacstallrepo.txt"
+	
+	fancy_message warn "There is no repo with the name $RED$REPONAME$NC"
```
wizard-28, line 70: img

```suggestion
	fancy_message warn "$RED$REPONAME$NC either doesn't exist or is not a pacstall programs repo"
```

I guess that makes more sense?

wizard-28 (changes requested):

misc/scripts/search.sh
D-Brox: I think "$REPONAME is not on your repo list or does not exist" would be better

wizard-28:
```suggestion
	fancy_message warn "$RED$REPONAME$NC is not on your repo list or doesn't exist"
```

Yeah.

wizard-28 (approved):

PR #273, opened 2021-10-01
fix(scripts): fix clutter on download

Henryws:
## Purpose

During upgrade, pacstall will dump pacscripts in the current dir, and when upgrading later on in the same directory, pacstall will attempt to install the pacscript, which would be the old one

## Approach

`cd` into `SRCDIR` on upgrade

### DIFF
--- a/misc/scripts/upgrade.sh
+++ b/misc/scripts/upgrade.sh
@@ -142,7 +142,7 @@ else
 	done < /tmp/pacstall-up-urls
 
 	export local='no'
-	cd $SRCDIR
+	cd "$SRCDIR"
 	for i in "${!upgrade[@]}"; do
 		PACKAGE=${upgrade[$i]}
 		ask "Do you want to upgrade ${GREEN}${PACKAGE}${NC}?" Y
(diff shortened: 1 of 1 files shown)


misc/scripts/upgrade.sh
```diff
@@ -142,6 +142,7 @@ else
 	done < /tmp/pacstall-up-urls
 
 	export local='no'
+	cd $SRCDIR
```
wizard-28, line 145:
```suggestion
	cd "$SRCDIR"
```

Fixes codacy issue

wizard-28: Also shouldn't we `cd` back to the user's original directory after this operation is complete?

wizard-28 (changes requested):

misc/scripts/upgrade.sh
Henryws: cd'ing in a script only works inside the script. What happens inside the script, stays in the script

D-Brox (approved):

PR #278, opened 2021-10-03
Autocompletions revamped for multirepo support

Henryws:
Missing flags were removing (-Qd) and -T was added. Autocompletions now
work for multiple repos

### DIFF
--- a/misc/completion/bash
+++ b/misc/completion/bash
@@ -31,22 +31,18 @@ _pacstall() {
 		;;
 
 		-S|--search)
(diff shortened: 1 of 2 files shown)


Henryws:
```bash
mapfile -t packages <<< "$(compgen -W "$(sed -e 's/$/\/packagelist/' /usr/share/pacstall/repo/pacstallrepo.txt | xargs -n 1 curl -s | awk '!seen[$0]++')" -- "$cur")"

_pacstall() {
	cur=${COMP_WORDS[COMP_CWORD]}
	prev=${COMP_WORDS[COMP_CWORD-1]}
	case "$prev" in
		-'?'|--help|-v|--version)
			return
		;;
```
this causes zsh completions to not work

Issue #284, opened 2021-10-07
Cannot remove package installed from local pacscript


Zahrun:
**Describe the bug**
Cannot remove package installed from local pacscript

**To Reproduce**
1. Install a package with `pacstall -Il X.pacscript`
2. The package is installed and we can see it with `pacstall -L`
3. Impossible to remove the package with `pacstall -R`
`[!] ERROR: X is not installed or not properly symlinked`
`[!] ERROR: Failed to remove X`

**Expected behavior**
Remove package installed from local pacscript is possible

**Screenshots**
![image](https://user-images.githubusercontent.com/10415894/136469249-0b1daff9-42ef-422b-bc31-90b35d47f01e.png)

**Desktop (please complete the following information):**

- KUbuntu 21.04
- Pacstall 1.6 develop

Henryws:
have you tested this on other packages?

Henryws:
I just tested it. It seems to work for me
```bash
henry@twilight ~ % ››› pacstall -R webapp-manager
[+] INFO: Removing symlinks
[+] INFO: Removing package
[+] INFO: You may want to remove gir1.2-xapp-1.0 python3 python3-bs4 python3-configobj python3-gi python3-setproctitle python3-tldextract xapps-common python3-pil
[+] INFO: Removing dummy package
(Reading database ... 285232 files and directories currently installed.)
Removing webapp-manager (1.1.5-1) ...
[+] INFO: Package removed successfully
```

PR #285, opened 2021-10-07
Fix permissions on cp (#284)

Elsie19:


Zahrun:
Same change is required in the case of deb install, line 419

Henryws:
> Same change is required in the case of deb install, line 419

😅

PR #298, opened 2021-10-12
fix(scripts): Error handle pacscript functions

Elsie19:
## Purpose

Currently if `prepare`, `build`, or `install` fails, pacstall will zoom right by and continue with a broken state

## Approach

Add guards to the functions to check if it failed

## Checklist


Henryws:
Um, I do have a guard on `postinst`

PR #304, opened 2021-10-14
rem(installer): `ask()`

D-Brox:
## Purpose

Ask function is only used once in both the installer and uninstaller and doesn't need to be as complex as in the main script

## Approach

Remove ask function and replace it with simpler code

## Checklist

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -59,63 +59,6 @@ if [[ ! -t 0 ]]; then
 	fancy_message warn "Reading input from pipe"
 fi
 
-function ask() {
-	local prompt default reply
-
-	if [[ ${2:-} = 'Y' ]]; then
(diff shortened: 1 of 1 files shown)


D-Brox:
> I actually use the "Reading input from pipe" feature in my bootstrapping script for automatic Pacstall installation. Also people like distro maintainers may use the -P flag for the same in here?

Pipe works, `-P` flag doesn't

wizard-28:
> > I actually use the "Reading input from pipe" feature in my bootstrapping script for automatic Pacstall installation. Also people like distro maintainers may use the -P flag for the same in here?
> 
> Pipe works, `-P` flag doesn't. But to be fair, `-P` didn't work before either

So piping works in the PR right?

PR #347, opened 2021-11-12
Exit on download fail

Henryws:
## Purpose

Exit early, exit fast

## Approach

Add a check to detect if the file failed to download

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -196,9 +196,15 @@ function ask() {
 function download() {
 	sudo rm -f "${1##*/}"
 	if command -v axel > /dev/null; then
-		axel -ao "${1##*/}" "$1"
+		if ! axel -ao "${1##*/}" "$1"; then
+			fancy_message error "Could not download file"
+			exit 1
+		fi
 	else
-		wget -q --show-progress --progress=bar:force "$1" 2>&1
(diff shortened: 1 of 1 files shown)


wizard-28:
I think this would be better to implement in the python version


Syncing these 3 branches is a nightmare...

Henryws:
> I think this would be better to implement in the python version
> 
> Syncing these 3 branches is a nightmare...

We can sort out the logistics later. This is important

wizard-28:
> > I think this would be better to implement in the python version
> > 
> > Syncing these 3 branches is a nightmare...
> 
> We can sort out the logistics later. This is important

This is not really *that* important. If it were then we would have already received a bug report about this.

Henryws:
Well I’ve had bugs related to this. When a download somehow fails, it will continue onwards, and while we have guards in place, it looks messy to continue on after a failed download

wizard-28:
> Well I’ve had bugs related to this. When a download somehow fails, it will continue onwards, and while we have guards in place, it looks messy to continue on after a failed download

"looks messy" doesn't make it an important bugfix. Also it's not like the user's data is being corrupted or anything. It just fails. That also rarely as the links tend to work most times...

PR #351, opened 2021-11-13
feat(pacstall): Add locking mechanism

wizard-28:
## Approach

Add a locking mechanism to the main script, also move the argument parser to it's own function as a result.

## Checklist

### DIFF
--- a/src/pacstall.py
+++ b/src/pacstall.py
@@ -140,7 +140,7 @@ if __name__ == "__main__":
 
     if args.install or args.remove or args.upgrade:
         lock_file = open("/var/lock/pacstall.lock", "w")
-        call("/usr/bin/sudo /usr/bin/chown root /var/lock/pacstall.lock", shell=True)
+        call(["/usr/bin/sudo", "/usr/bin/chown", "root", "/var/lock/pacstall.lock"], shell=False)
         while True:
             try:
                 lockf(lock_file, LOCK_EX | LOCK_NB)
--- a/src/pacstall.py
+++ b/src/pacstall.py
@@ -35,7 +35,8 @@ from api import message
 
 # Copied from https://stackoverflow.com/a/23941599 and modified
 class CustomHelpFormatter(HelpFormatter):
-    """Custom help message formatter for Pacstall.
+    """
+    Custom help message formatter for Pacstall.
 
     Format:
     -s, --long       help message
@@ -63,7 +64,8 @@ class CustomHelpFormatter(HelpFormatter):
 
 
 def parse_arguments() -> Namespace:
-    """Parses command line arguments passed to Pacstall.
+    """
+    Parses command line arguments passed to Pacstall.
 
     Prints help and exits if no argument is passed.
 
--- a/src/pacstall.py
+++ b/src/pacstall.py
@@ -140,7 +140,7 @@ if __name__ == "__main__":
 
     if args.install or args.remove or args.upgrade:
         lock_file = open("/var/lock/pacstall.lock", "w")
-        call(["/usr/bin/sudo", "/usr/bin/chown", "root", "/var/lock/pacstall.lock"], shell=False)
+        call(["/usr/bin/sudo", "/usr/bin/chown", "root", "/var/lock/pacstall.lock"])
         while True:
             try:
                 lockf(lock_file, LOCK_EX | LOCK_NB)
--- a/src/pacstall.py
+++ b/src/pacstall.py
@@ -133,7 +133,7 @@ def parse_arguments() -> Namespace:
 
 if __name__ == "__main__":
     if getuser() == "root":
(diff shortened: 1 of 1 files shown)


src/pacstall.py
```diff
@@ -59,61 +62,88 @@ def _format_action_invocation(self, action) -> str:
...
+    if len(sys.argv) == 1:
+        parser.print_help()
+        sys.exit(1)
+
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    if getuser() == "root":
+        message.fancy("error", "Pacstall can't be run as root")
```
Henryws, line 134: No contractions: `Pacstall cannot be run as root`

src/pacstall.py
```diff
@@ -59,61 +62,88 @@ def _format_action_invocation(self, action) -> str:
...
+    if args.install or args.remove or args.upgrade:
+        global lock_file
+        lock_file = open("/var/lock/pacstall.lock", "w")
+        call("/usr/bin/sudo /usr/bin/chown root /var/lock/pacstall.lock", shell=True)
+        while True:
+            try:
+                lockf(lock_file, LOCK_EX | LOCK_NB)
+                break
+            except IOError:
+                message.fancy("error", "Another instance running lol")
```
Henryws, line 148: Change it

wizard-28: Oof. Right. Can you use the suggest button and suggest the change directly?

src/pacstall.py
wizard-28: Okay. But you should have suggested this change when I merged the PR related to this... Also use the suggest button.

src/pacstall.py
Henryws: Next time I will

wizard-28: You know you can do it right now right?

wizard-28: It appears on the ribbon when you start typing the comment here.

src/pacstall.py
Henryws, line 134: `message.fancy("error", "Pacstall cannot be run as root")`

src/pacstall.py
Henryws: `message.fancy("error", "Another instance is already running")`

Henryws (changes requested):

src/pacstall.py
wizard-28: That's not a suggestion... Click the button (or <kbd>Ctrl</kbd> + <kbd>g</kbd>  alternatively) then write your code there.

src/pacstall.py
wizard-28: That's not a suggestion... Click the button (or <kbd>Ctrl</kbd> + <kbd>g</kbd>  alternatively) then write your code there.

src/pacstall.py
```diff
@@ -59,61 +63,88 @@ def _format_action_invocation(self, action) -> str:
...
+    if len(sys.argv) == 1:
+        parser.print_help()
+        sys.exit(1)
+
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    if getuser() == "root":
+        message.fancy("error", "Pacstall can't be run as root")
```
Henryws, line 136:
```suggestion
        message.fancy("error", "Pacstall cannot be run as root")
```

src/pacstall.py
```diff
@@ -59,61 +63,88 @@ def _format_action_invocation(self, action) -> str:
...
+
+    if args.install or args.remove or args.upgrade:
+        lock_file = open("/var/lock/pacstall.lock", "w")
+        call("/usr/bin/sudo /usr/bin/chown root /var/lock/pacstall.lock", shell=True)
+        while True:
+            try:
+                lockf(lock_file, LOCK_EX | LOCK_NB)
+                break
+            except IOError:
+                message.fancy("error", "Another instance running lol")
```
Henryws, line 149:
```suggestion
                message.fancy("error", "Another instance running lol")
```

Henryws (changes requested):

src/pacstall.py
wizard-28:
```suggestion
                message.fancy("error", "Pacstall is already running another instance")
```

How about this? This message was used in the bash version

Henryws: that’s good

Henryws (approved):

D-Brox (approved):
All is right, tho I feel like the lock mechanism should be in the install and remove modules, to avoid them being run independently if we imported them in a custom python script

wizard-28:
> All is right, tho I feel like the lock mechanism should be in the install and remove modules, to avoid them being run independently if we imported them in a custom python script

Hmm.... I guess we could always do that but can you give a situation when we'll import then elsewhere?

PR #353, opened 2021-11-17
fix(install): Don't split on `-` while checking `breaks`

D-Brox:
## Purpose

Fix `$breaks` check when re-installing or updating a package

## Approach

Replace check's regrex to not split on `-`, only on spaces

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -327,7 +327,7 @@ if [[ -n "$breaks" ]]; then
 			cleanup
 			return 1
 		fi
-		if [[ "${pkg}" != "${name}" ]] && pacstall -L | grep "${pkg}" > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


Henryws (approved):

Henryws:
@wizard-28 can you review this bugfix?

wizard-28:
> Purpose

> Fix $breaks check when re-installing or updating a package

How was it broken?

wizard-28:
Can you give an example pacscript where this bug caused issues before?

Henryws:
[Bemenu](https://github.com/pacstall/pacstall-programs/blob/master/packages/bemenu-git/bemenu-git.pacscript), because it installs to `bemenu`, but in `breaks`, it breaks `bemenu`

D-Brox:
@wizard-28 updating any `pkgname-git` if `pkgname` is in `breaks`. I found this issue with `fastfetch-git`

wizard-28 (approved):

PR #403, opened 2021-12-07
fix(installer): missing directory permissions on installation

Henryws:
## Purpose

During install, the directories `/tmp/pacstall` and `/var/log/pacstall` do not have the proper permissions, which means on a clean install of pacstall, it refuses to install anything

## Approach

Correct permissions during install of pacstall

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -134,4 +134,5 @@ wait
 
 chmod +x "/bin/pacstall"
 chmod +x $STGDIR/scripts/*
+sudo chown -R "$USER":"$USER" {/var/log/pacstall/,/tmp/pacstall/}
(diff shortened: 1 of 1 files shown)


wizard-28 (changes requested):
```bash
user="$(logname 2> /dev/null)" || "${SUDO_USER:-${USER}}"
```

Code snippet made with the help of @louis77 

I guess we should refactor the whole code base using this?

*P.S: Comment first before making any changes. Also test the above snippet*

Henryws:
Yeah, we should refractor, and use that code snippet

wizard-28:
@Henryws, should this be closed then?

PR #405, opened 2021-12-09
fix: inconsistent user account detection

Henryws:
## Purpose

When logging into some systems with a display manager, the `logname` command or `$USER` break sometimes due to unset variables usually handled by the DM.

## Approach

Replace all instances of the previous commands with a single variable that tests for `logname` first (most reliable), then tries `$SUDO_USER`, then `$USER` finally.

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -31,7 +31,7 @@ export SRCDIR="/tmp/pacstall"
 export STGDIR="/usr/share/pacstall"
 export STOWDIR="/usr/src/pacstall"
 
-export user=$(logname 2> /dev/null || echo "${SUDO_USER:-${USER}}")
+export PACSTALL_USER=$(logname 2> /dev/null || echo "${SUDO_USER:-${USER}}")
 
 # Colors
 BOLD=$(tput bold)
--- a/pacstall
+++ b/pacstall
@@ -31,6 +31,8 @@ export SRCDIR="/tmp/pacstall"
 export STGDIR="/usr/share/pacstall"
 export STOWDIR="/usr/src/pacstall"
 
+export user=$(logname 2> /dev/null || echo "${SUDO_USER:-${USER}}")
+
 # Colors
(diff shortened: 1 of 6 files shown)


pacstall
```diff
@@ -31,6 +31,8 @@ export SRCDIR="/tmp/pacstall"
 export STGDIR="/usr/share/pacstall"
 export STOWDIR="/usr/src/pacstall"
 
+export user=$(logname 2> /dev/null || echo "${SUDO_USER:-${USER}}")
```
wizard-28, line 34: Shouldn't it be capitalized and something else (so as to not conflict with `USER`), it's a constant after all.

wizard-28 (changes requested):

wizard-28:
> When logging into some systems, the logname command or $USER break sometimes

This description is embarrassing (due to being *so* vague). File a bug report and link it to this PR.

wizard-28 (changes requested):
fix the one in the installer, and the uninstaller

wizard-28 (changes requested):
You missed `upgrade.sh`

Henryws:
> You missed `upgrade.sh`

Where? I can't find any reference of `$USER` or `logname`

wizard-28:
> > You missed `upgrade.sh`
> 
> Where? I can't find any reference of `$USER` or `logname`

`$LOGNAME`

Henryws:
> You missed `upgrade.sh`

I think you meant `update.sh`

wizard-28 (changes requested):
You missed `error_log.sh`

wizard-28 (approved):
Looks good, but didn't test SSH (which is the main reason this change is being done). Hoping @D-Brox would do it.

D-Brox (approved):
Tested through ssh, worked perfectly

PR #426, opened 2021-12-28
fix(pacstall): outdated version printer for scripting

Henryws:
## Purpose

The version printer `pacstall -V $package` is outdated and was never fixed for newer version of Pacstall leading to weird and breaking output when run.

## Approach

Rework the system

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -458,9 +458,8 @@ while [[ ! "$1" == "--" ]]; do
 				echo -e "1.7 \033[1m\x1b[38;2;160;82;45mSienna${NC}"
 				exit 0
 			else
-				# If pacstall -V was called with an argument, it's a package, so get the package version (usful for scripting)
 				if ! source "$LOGDIR/$PACKAGE"; then
-					fancy_message error "${GREEN}${PACKAGE}${NC} ins not installed"
+					fancy_message error "${GREEN}${PACKAGE}${NC} is not installed"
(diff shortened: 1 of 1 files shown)


pacstall
```diff
@@ -459,11 +459,12 @@ while [[ ! "$1" == "--" ]]; do
 				exit 0
 			else
 				# If pacstall -V was called with an argument, it's a package, so get the package version (usful for scripting)
```
wizard-28, line 461: What about this comment?

pacstall
```diff
@@ -459,11 +459,12 @@ while [[ ! "$1" == "--" ]]; do
...
-				if pacstall -Qi "$PACKAGE" | grep 'version' | cut -d ":" -f2 | sed 's/ //g'; then
-					exit 0
-				else
+				if ! source "$LOGDIR/$PACKAGE"; then
 					fancy_message error "${GREEN}${PACKAGE}${NC} ins not installed"
```
wizard-28, line 463: What is `ins`

wizard-28 (changes requested):
Make a bug report and link it here.

pacstall
Henryws: Typo, oops

wizard-28 (commented):
What about the completions though? Since this was depreciated we removed the completions for this.

PR #436, opened 2022-01-01
feat: allow prepare function to be absent

Henryws:
## Purpose

This will allow pacscripts to run without a `prepare` function, so we don't have to put `true` in them.

## Approach

Add a check to `prepare` to make sure it exists, and if not, skip it.

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -507,18 +507,20 @@ fi
 
 export pkgdir="/usr/src/pacstall/$name"
 
-fancy_message info "Preparing"
-if ! prepare; then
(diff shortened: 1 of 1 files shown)


D-Brox (approved):

wizard-28:
Eh, not now. Three problems with this
1. This is a depredation, deprecation can only take place in feature releases.
2. If you make `prepare` optional, then you have to make `build` and `install` optional too.
3. You didn't make a feature request for this. So we could probably have discussed this earlier in the issue instead of wasting time making this PR, which now will be closed. (I've told you to do so probably 10 times already, it's like speaking to a brick wall)

PR #437, opened 2022-01-01
fix(remove): removal of unidentified packages

Henryws:
## Purpose

`return` does not exit removal if a package doesn't exist, but `exit` does.

## Approach

Replace `return` with `exit`

## Checklist


wizard-28 (commented):
What about the `return 1` in L107? Shouldn't that be an `exit 1` too?

D-Brox (approved):

wizard-28 (approved):

PR #439, opened 2022-01-01
fix: miscellaneous possible bugs due to possible dashes in inputs

Henryws:
## Purpose

To prevent bugs because of malformed input, like a dash, which would be evaluated by bash as a flag.

## Approach

Add `--` before inputs so bash evaluates them as inputs instead of flags

## Checklist

### DIFF
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -50,7 +50,7 @@ else
 fi
 
 
-if ! wget -q --spider "$REPO/packagelist"; then
+if ! wget -q --spider -- "$REPO/packagelist"; then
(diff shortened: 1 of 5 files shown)


wizard-28 (commented):
Are you sure these are the only lines vulnerable to this kind of bug?

Henryws:
There are more places where we *could* put them, but these are the places where a dash is possible. The ones I fixed are where the input is fluid and could contain a dash. The rest are fixed to a certain string.

wizard-28 (approved):
All worked, but didn't test `add-repo`

Henryws:
> All worked, but didn't test `add-repo`

```bash
[*] WARNING: The repo link must be the root to the raw files
[*] WARNING: Make sure the repo contains a package list
Do you want to add "-" to the repo list? [y/N] y
[*] WARNING: If the URL is a private repo, edit /usr/share/pacstall/repo/pacstallrepo.txt
[!] ERROR: packagelist file not found
```

D-Brox (approved):

Issue #443, opened 2022-01-07
Pacstall should give a better error message when `gives` contains a space


Henryws:
**Describe the bug**
`dpkg` does not allow spaces in `Package`, and our dummy deb maker does not prevent this from happening, which results in a very vague error message.

**To Reproduce**
Install Pacstall
Edit a package to have a space, or two strings in `gives`
Observe error

**Expected behavior**
For a clear error of what happened, perhaps by checking for a space before creating the deb, inside the checking function.

**Screenshots**
![image](https://user-images.githubusercontent.com/58742515/148470866-149db623-8b42-4de7-81e0-32d384d1d78e.png)

**Desktop (please complete the following information):**

- OS + version: Debian Sid
- Pacstall Version + branch: 1.7 Master

**Additional context**
https://stackoverflow.com/a/1474094/13449010
https://github.com/pacstall/pacstall-programs/issues/622

PR #444, opened 2022-01-08
fix(install): ambiguous error message when `$gives` contains multiple fields

paulcosma97:
Logs an error message when `$gives` contains multiple words by checking for whitespaces and commas.

## Purpose

Fixes issue #443.

## Approach

Checks if `$gives` contains whitespace or commas.

## Progress

- [x] Add check condition
- [x] log the error and `exit 1` if multiple words are found.

## Addendum

Issue #443 

## Checklist

- [x] ~I will not be sued if I forgot something.~ I approve to be punished on the rack, if I falter in abiding by the clauses!

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -68,7 +68,7 @@ function checks() {
 		exit 1
 	fi
 	if echo "$gives" | grep -q ",\|\\s"; then
-		fancy_message error "Expected argument \"gives\" to only contain one word but has multiple"
+		fancy_message error "\"gives\" supports only one field"
 		exit 1
 	fi
 	if [[ -z "$url" ]]; then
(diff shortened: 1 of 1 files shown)


misc/scripts/install-local.sh
```diff
@@ -67,6 +67,10 @@ function checks() {
 		fancy_message error "Package does not contain version"
 		exit 1
 	fi
+	if echo "$gives" | grep -q ",\|\\s"; then
+		fancy_message error "Expected argument \"gives\" to only contain one word but has multiple"
```
wizard-28, line 71:
```suggestion
		fancy_message error "\`gives\` supports only one field"
```

How about this?

wizard-28 (changes requested):

misc/scripts/install-local.sh
Henryws: We don't use backticks anywhere else in the codebase. I think we should keep it how it is but use @wizard-28 last part (`supports only one field`)

paulcosma97: aight done

Henryws (approved):

wizard-28 (approved):

PR #455, opened 2022-01-10
feat(main)!: only launch as root

### LINKED ISSUE #454: Launch `pacstall` using root
We should run the `pacstall` command using root, as most of the operations require root privileges.

We could later downscale privileges.

wizard-28:
BREAKING CHANGE: The program can only be launched as root.

## Purpose

Closes #454 

## Checklist

### DIFF
--- a/pacstall/__main__.py
+++ b/pacstall/__main__.py
@@ -32,6 +32,7 @@ from time import sleep
 from rich.traceback import install
 
 from pacstall.api import message
+from pacstall.api.error_codes import ErrorCodes
 from pacstall.cmds import download
 from pacstall.parser import parse_arguments
 
@@ -45,7 +46,7 @@ if __name__ == "__main__":
 
     if getuser() != "root":
         message.fancy("error", "Pacstall needs to be launched as root!")
-        sys.exit(64)  # --> command line usage error
+        sys.exit(ErrorCodes.USAGE_ERROR)  # --> command line usage error
 
     if args.command in ["install", "remove", "upgrade"]:
         lock_file = open("/var/lock/pacstall.lock", "w")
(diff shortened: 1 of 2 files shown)


Henryws (approved):

pacstall/__main__.py
```diff
@@ -41,16 +40,15 @@
...
-    if getuser() == "root":
-        message.fancy("error", "Pacstall cannot be run as root")
-        sys.exit(1)
-
     args = parse_arguments()
     print(args)
 
+    if getuser() != "root":
+        message.fancy("error", "Pacstall needs to be launched as root!")
+        sys.exit(64)  # --> command line usage error
```
paulcosma97, line 48: If we're going to be using a set of standard error codes, we should either stick to the Linux Standard or create our own and reference them by a common enum

According to [this article](https://www.baeldung.com/linux/status-codes) code `126` would be better.

paulcosma97: IMO we should have our own values hardcoded in some kind of enum and use that

wizard-28: > If we're going to be using a set of standard error codes, we should either stick to the Linux Standard or create our own and reference them by a common enum
> 
> According to [this article](https://www.baeldung.com/linux/status-codes) code `126` would be better.

I got `64` from the [`sysexits.h`](https://www.freebsd.org/cgi/man.cgi?query=sysexits) (this is a standard UNIX header). From what I can see that article (it's exit codes) is for bash scripts (https://stackoverflow.com/a/1101969).

wizard-28: > IMO we should have our own values hardcoded in some kind of enum and use that

Yeah, we could use that for specific error codes, but I think for common errors like this one we should follow the standard error codes.

Do you want me to add an Enum of all the error codes (the standard ones for now) in `pacstall/api/error_codes.py`?

paulcosma97: Yup, that sounds good

wizard-28: @paulcosma97 done

paulcosma97 (approved):

Issue #340, opened 2021-11-09
NEWS: Ongoing Rust rewrite


wizard-28:

wizard-28:
Also we're changing to `GPL-3.0-or-later` from `GPL-3.0-only` in this rewrite.

wizard-28:
We're currently working on the [LibPacstall](https://github.com/pacstall/libpacstall) back-end library, which is written in Rust.

Python bindings will be developed after it releases, and we'll use that in Pacstall

wizard-28:
We've decided to go full Rust for the rewrite.

Slackadays:
@Elsie19 have you considered Ada for the rewrite? It is super-duper safe and would be really cool to use

Elsie19:
So the rewrite (in any language) is looking more and more like a pipe dream and we're probably going to stick with Bash and probably do heavy refactoring.

Elsie19:
😈

Elsie19:
https://github.com/Elsie19/libpacstall/

PR #485, opened 2022-01-31
some bash syntax managment

zakariaGatter:
first of all when i edit your script i don't try to say that you do a bad job all im trying to do is to make your script better cause i like what you doing 

all support please continue

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -151,43 +151,39 @@ function ask() {
 	fi
 
 	# Ask the question (not using "read -p" as it uses stderr not stdout)
-	echo -ne "$1 [$prompt] "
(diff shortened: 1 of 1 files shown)


wizard-28:
Thanks for making this pull request, but we are rewriting Pacstall in Python (for the most part, some part will still be in bash) for the next version (see https://github.com/pacstall/pacstall/issues/340, and [our project tracker](https://github.com/orgs/pacstall/projects/2/views/6)). No more commits would be made on the `develop` branch for now, `python3-rewrite`'s the active branch. We would appreciate it if you helped out with the rewrite.

PR #488, opened 2022-02-02
some cmd fixes and syntax optimizing

zakariaGatter:
try my best to replace some repetitive actione and optimize some command for better performance

* take `wc` when count file line number using only `awk`
* use a loop to create and delete tmp files
* make arrays from files with a builtin bash function `mapfile`
* ....

thank you and i hope this changes make your script better

### DIFF
--- a/src/scripts/upgrade.sh
+++ b/src/scripts/upgrade.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 #     ____                  __        ____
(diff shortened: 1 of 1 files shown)


Henryws:
We are currently working on the Python rewrite, which means no more bash code should be altered.

Henryws:
If you want to, you can join our discord/matrix, because the Python rewrite will contain a bash script that does the actual installing of the package.

PR #496, opened 2022-02-07
fix: allow certain commands to run with root

Henryws:
## Purpose

Prevent certain flags from running without root, and offer a suggestion to run if it fails.

## Approach

Check if the user is root.

## Checklist


wizard-28 (changes requested):

wizard-28:
@Henryws run the pre-commit hooks...............

wizard-28 (approved):

PR #498, opened 2022-02-11
docs(readme): replace future tense word with definitive

Henryws:


wizard-28 (approved):
Shame this took so long.

PR #503, opened 2022-02-13
refractor: differentiate between file existence and network errors

Henryws:
## Purpose

Shows the difference between a broken link, or no network.

## Approach

Curl the header first, and check the error code.

## Addendum

Fixes #502

## Checklist

- [x] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

PR #504, opened 2022-02-17
fix(search): fail out on no package

Henryws:
## Purpose

Currently `search.sh` uses `return` instead of `exit` to signify a non existent package, but that doesn't work, which gives an exit code of zero instead of one.

## Approach

Replace certain `return`s with `exit`s

## Checklist

### DIFF
--- a/misc/scripts/search.sh
+++ b/misc/scripts/search.sh
@@ -57,7 +57,7 @@ if [[ $PACKAGE == *@* ]]; then
 			if [[ "$LEN" -eq 0 ]]; then
 				fancy_message warn "There is no package with the name $IRed${PACKAGE%%@*}$NC in the repo $CYAN$REPONAME$NC"
 				error_log 3 "search $PACKAGE@$REPONAME"
-				return 1	
+				exit 1
 			fi
 			export PACKAGE
 			export REPO="$URL"
@@ -67,7 +67,7 @@ if [[ $PACKAGE == *@* ]]; then
 	
 	fancy_message warn "$IRed$REPONAME$NC is not on your repo list or does not exist"
 	error_log 3 "search $PACKAGE@$REPONAME"
-	return 1	
+	exit 1
 fi
 
 # Makes array of packages and array
--- a/misc/scripts/search.sh
+++ b/misc/scripts/search.sh
@@ -114,8 +114,10 @@ if [[ "$LEN" -eq 0 ]]; then
 	if [[ -z "$SEARCH" ]]; then
 		fancy_message warn "There is no package with the name $IRed$PACKAGE$NC"
 		error_log 3 "search $PACKAGE"
+		exit 1
 	else
 		fancy_message warn "There is no package with the name $IRed$SEARCH$NC"
+		exit 1
 	fi
 
 	return 1


wizard-28:
I think we've discussed this 2 times prior, and I seem to have forgotten why we went with return.

Henryws:
`return` is used for functions to give a return value, and `exit` exits a script. A `return` will do nothing if not in a function, which this PR fixes, by changing the offending `return`s to `exit`s, so pacstall errors out with the correct exit code

misc/scripts/search.sh
```diff
@@ -57,7 +57,7 @@ if [[ $PACKAGE == *@* ]]; then
 			if [[ "$LEN" -eq 0 ]]; then
 				fancy_message warn "There is no package with the name $IRed${PACKAGE%%@*}$NC in the repo $CYAN$REPONAME$NC"
 				error_log 3 "search $PACKAGE@$REPONAME"
-				return 1	
+				exit 1
```
D-Brox, line 60: When installing multiple packages at the same time, or when upgrading, I think this causes pacstall to stop  instead of continuing to the next package on the list

D-Brox, line 60: When installing multiple packages at the same time, or when upgrading, I think this causes pacstall to stop  instead of continuing to the next package on the list

misc/scripts/search.sh
```diff
@@ -67,7 +67,7 @@ if [[ $PACKAGE == *@* ]]; then
 	
 	fancy_message warn "$IRed$REPONAME$NC is not on your repo list or does not exist"
 	error_log 3 "search $PACKAGE@$REPONAME"
-	return 1	
+	exit 1
```
D-Brox, line 70: Same thing here

misc/scripts/search.sh
```diff
@@ -114,8 +114,10 @@ if [[ "$LEN" -eq 0 ]]; then
 	if [[ -z "$SEARCH" ]]; then
 		fancy_message warn "There is no package with the name $IRed$PACKAGE$NC"
 		error_log 3 "search $PACKAGE"
+		exit 1
```
D-Brox, line 117: This one is for install or upgrade, probably needs changing too

misc/scripts/search.sh
```diff
@@ -114,8 +114,10 @@ if [[ "$LEN" -eq 0 ]]; then
...
 	else
 		fancy_message warn "There is no package with the name $IRed$SEARCH$NC"
+		exit 1
```
D-Brox, line 120: This one is for search, so keep it as exit

misc/scripts/search.sh
Henryws: image

D-Brox: Ok, then error out on install. But what about on upgrade tho?

Henryws: That should be fixed with #503

wizard-28:
@Henryws this is going stale.

Henryws:
Give me one more week to finish this PR.

misc/scripts/search.sh
Henryws: There is no function this is in, so having return is wrong.

Henryws:
@wizard-28 ready.

D-Brox (approved):

wizard-28 (approved):

PR #506, opened 2022-02-19
fix(install-local): allow apt to downgrade packages

Henryws:
## Purpose

When installing the dummy deb, a git commit id can mess with the versioning, and some commits will be seen as a downgrade to apt.

## Approach

Allow downgrades

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -242,7 +242,7 @@ fi' | sudo tee "$SRCDIR/$name-pacstall/DEBIAN/postrm" >"/dev/null"
 	fi
 	export PACSTALL_INSTALL=1
 	sudo rm -rf "$SRCDIR/$name-pacstall"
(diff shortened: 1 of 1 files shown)


wizard-28 (commented):
Won't this affect pinned dummy debs or something? I forgot what we exactly used.

Henryws:
> Won't this affect pinned dummy debs or something? I forgot what we exactly used.

image
It doesn't effect dummy deb pinning

wizard-28:
> > Won't this affect pinned dummy debs or something? I forgot what we exactly used.
> 
> image It doesn't effect dummy deb pinning

Isn't there a better way to do this?

Henryws:
Not until the rewrite

wizard-28 (changes requested):
Write a FAQ, and add a comment in the code.

wizard-28 (approved):
Okay, but @D-Brox's approval needed

D-Brox (approved):

PR #523, opened 2022-03-18
fix(install-local): skip `$breaks` check for  `-deb` upgrades

D-Brox:
## Purpose

Fix for the upgrade command of some `-deb` packages installed through pacstall, which were unable to upgrade due to the `$breaks` variable

## Approach

Skip $breaks check if package already installed

## Checklist


wizard-28 (approved):
@Henryws

Henryws (approved):

PR #531, opened 2022-03-24
fix(install-local): properly display and install optional dependencies

D-Brox:
## Purpose

Currently optdepends do not get installed by apt due to the replacement of `dpgk` for `apt` when installing in #491.
Pacstall also asks if you want to install those dependencies even if you have them installed.

## Approach

Rework dummy deb installation and filter already installed optdeps

## Checklist


D-Brox:
Before you ask, Wizard, I just rebased it to fix a commit message : )

wizard-28 (changes requested):
Fix codacy issues.

wizard-28 (approved):

misc/scripts/install-local.sh
```diff
@@ -160,29 +160,53 @@ function makeVirtualDeb {
...
+		if [[ -n $optdeps ]]; then
+			fancy_message info "$name has optional dependencies that can enhance its functionalities"
+			echo "Optional dependencies:"
+			printf '    %s\n' "${optdeps[@]}"
+			ask "Do you want to install them" Y
+			if [[ $answer -eq 1 ]]; then
+				for optdep in "${optdeps[@]}"; do
+					depends+=" ${optdep%%: *}"
+				done
+				if pacstall -L | grep "$name" > /dev/null 2>&1; then
```
Henryws, line 194: This grep should match the entire line, not just a match, because a package could also have a `-git`, which would also be matched.

Henryws (changes requested):

misc/scripts/install-local.sh
D-Brox: True, there is even the fixed version somewhere in the code

misc/scripts/install-local.sh
```diff
@@ -160,29 +160,53 @@ function makeVirtualDeb {
...
+			if [[ $answer -eq 1 ]]; then
+				for optdep in "${optdeps[@]}"; do
+					depends+=" ${optdep%%: *}"
+				done
+				if pacstall -L | grep -E "(^| )${name}( |$)"> /dev/null 2>&1; then
+					sudo dpkg -r --force-all "$name" > /dev/null
+				fi
+			else
+				# Add to the suggests anyway. They won't get installed but can be queried
+				printf "Suggests:" |sudo tee -a "$SRCDIR/$name-pacstall/DEBIAN/control" > /dev/null
```
Henryws, line 199:
```suggestion
				printf "Suggests:" | sudo tee -a "$SRCDIR/$name-pacstall/DEBIAN/control" > /dev/null
```

Henryws (changes requested):

Henryws (approved):

wizard-28 (commented):
What about the Codacy issues @D-Brox?

PR #534, opened 2022-03-26
fix(check_url): use a faster method of checking http codes

Henryws:
## Purpose

The current method in `check_url` downloads the file, but redirects to `/dev/null`, meaning that the file is downloaded multiple times, which is very bad on slow connections, as well as general bandwidth.

## Approach

curl the headers, which will lead to significantly less download times

## Checklist


wizard-28:
I assume you're still working on this, as this is marked as a draft.

D-Brox (approved):

D-Brox (changes requested):
Remove `check_url ` from `checks` function. Add return code check on `download` calls on the switch case

PR #535, opened 2022-03-27
 fix: add guards to download and speed up url checking 

Henryws:
## Purpose

Currently, in the `develop` branch, the system for determining if a URL was missing, broken, or working is very inefficient, downloading the same file in many different places. 

## Approach

Combine `check_url` and `download`, and make `curl` only check the head of a URL, meaning it won't download the whole file to check the status.

## Checklist


misc/scripts/install-local.sh
```diff
@@ -465,7 +462,12 @@ if [[ "$name" == *-git ]]; then
 else
 	case "$url" in
 		*.zip)
-			download "$url"
+			if ! download "$url"; then
+				error_log 1 "download $PACKAGE"
```
D-Brox, line 466: Add `fancy_message error`.
Same with the others

D-Brox (changes requested):

D-Brox (approved):

wizard-28 (approved):

PR #582, opened 2022-05-04
fix(install-local): skip `$replace` check if package is already installed

D-Brox:
## Purpose

We don't need to check `$replaces` if the package is already installed. This is happening during reinstall and upgrade.

## Checklist


Henryws (approved):

wizard-28 (changes requested):
Run pre-commit hooks

D-Brox:
> Run pre-commit hooks

I ran it locally, the only "issues" found is with old code (only warnings)

PR #600, opened 2022-06-10
fix(install): faulty network check

wizard-28:
## Purpose

Fixes faulty network checks in `install.sh`

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -51,6 +51,28 @@ function fancy_message() {
 	esac
 }
 
+function check_url() {
+	http_code="$(curl -o /dev/null -s --head --write-out '%{http_code}\n' -- "${1}")"
+	case "${http_code}" in
+		000)
+			fancy_message error "Failed to download file, check your connection"
+			error_log 1 "get ${PACKAGE} pacscript"
+			return 1
+			;;
+		404)
+			fancy_message error "The URL cannot be found"
+			return 1
+			;;
+		200 | 301 | 302)
+			true
+			;;
+		*)
+			fancy_message error "Failed with http code ${http_code}"
+			return 1
+			;;
+	esac
+}
+
 if [[ ! -t 0 ]]; then
 	NON_INTERACTIVE=true
 	fancy_message warn "Reading input from pipe"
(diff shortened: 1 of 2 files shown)


D-Brox (approved):

uninstall.sh
```diff
@@ -46,10 +46,10 @@ function fancy_message() {
 	local MESSAGE="${2}"
 
 	case ${MESSAGE_TYPE} in
-		info) echo -e "[${GREEN}+${NC}] INFO: ${MESSAGE}" ;;
-		warn) echo -e "[${YELLOW}*${NC}] WARNING: ${MESSAGE}" ;;
-		error) echo -e "[${RED}!${NC}] ERROR: ${MESSAGE}" ;;
-		*) echo -e "[?] UNKNOWN: ${MESSAGE}" ;;
+		info) echo -e "[${BGreen}+${NC}] INFO: ${MESSAGE}" ;;
```
D-Brox, line 49: Fix variable names for the colors color

D-Brox (changes requested):

install.sh
```diff
@@ -67,10 +89,10 @@ echo -e "|---${GREEN}Pacstall Installer${NC}---|"
 echo -e "|------------------------|"
 
 if [[ ${GITHUB_ACTIONS} != "true" ]]; then
-	if ! ping -c 1 github.com &> /dev/null; then
-		fancy_message error "You seem to be offline"
+	check_url "https://github.com" || check_url "https://gitlab.com" || {
```
Henryws, line 92: This should only be to GitHub, as that's where we pull scripts from, so if githubs down but gitlab works, it doesn't matter

Henryws (changes requested):

install.sh
wizard-28: We support custom repos from gitlab

Henryws: We do, but our scripts are located on GitHub at the moment, meaning that this ping to gitlab doesn't matter, because if githubs down and gitlab isn't, we still can't pull scripts

wizard-28: Well, what if the user has removed the official repo and is only using a gitlab custom repo

uninstall.sh
D-Brox: They aren't defined in this script tho
image

D-Brox (changes requested):
Fix variable names
https://github.com/pacstall/pacstall/pull/600#discussion_r912401712

D-Brox (approved):

Henryws (approved):

PR #601, opened 2022-06-12
feat!: support for local repos

D-Brox:
## Purpose

Add support for locally hosted repos

## Approach

Use `curl` instead of `wget` due to it's  `file://` protocol support. Format path  in `add-repo` when adding local repos.

## Checklist

### DIFF
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -53,7 +53,7 @@ else
 	fi
 fi
 
-if ! curl --head -s --fail -- "$REPO/packagelist"; then
+if ! curl -s --fail -- "$REPO/packagelist"; then
 	fancy_message warn "If the URL is a private repo, edit ${CYAN}\e]8;;file://$STGDIR/repo/pacstallrepo.txt\a$STGDIR/repo/pacstallrepo.txt\e]8;;\a${NC}"
 	fancy_message error "packagelist file not found"
 	exit 3
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -39,6 +39,10 @@ elif echo "$REPO" | grep "gitlab.com" > /dev/null; then
 	else
 		REPO="${REPO/"/tree/"/"/raw/"}"
 	fi
+elif [[ -d "$REPO" ]] > /dev/null; then
+	if ! echo "$REPO" | grep "file://" > /dev/null; then
+		REPO="file://$(readlink -f $REPO)"
+	fi
 else
 	fancy_message warn "The repo link must be the root to the raw files"
 	fancy_message warn "Make sure the repo contains a package list"
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -53,7 +53,7 @@ else
 	fi
 fi
 
-if ! curl -s --fail -- "$REPO/packagelist"; then
+if ! curl --head -s --fail -- "$REPO/packagelist" > /dev/null; then
 	fancy_message warn "If the URL is a private repo, edit ${CYAN}\e]8;;file://$STGDIR/repo/pacstallrepo.txt\a$STGDIR/repo/pacstallrepo.txt\e]8;;\a${NC}"
 	fancy_message error "packagelist file not found"
 	exit 3
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -39,7 +39,7 @@ elif echo "$REPO" | grep "gitlab.com" > /dev/null; then
 	else
 		REPO="${REPO/"/tree/"/"/raw/"}"
 	fi
-elif [[ -d "$REPO" ]] > /dev/null; then
+elif [[ -d $REPO ]] > /dev/null; then
 	if ! echo "$REPO" | grep "file://" > /dev/null; then
 		REPO="file://$(readlink -f $REPO)"
 	fi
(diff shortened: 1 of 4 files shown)


Henryws (changes requested):
If a secondary repo is misconfigured, operations like install and download fail.

Make sure to allow properly configured repos to work even if a misconfigured one exists. Add a warning about the misconfigured one

pacstall
```diff
@@ -194,25 +194,35 @@ function ask() {
...
-			fancy_message error "Failed with http code ${http_code}"
-			return 1
-			;;
-	esac
+    if [[ "${1}" == "file://"* ]]; then
+    	if [[ -f "${1}" ]]; then
+    	  true
+    	fi
+    else
+		http_code="$(curl -o /dev/null -s --head --write-out '%{http_code}\n' -- "${1}")"
```
Henryws, line 202: This doesn't have the same tabbing as lines above

Henryws (changes requested):

pacstall
D-Brox: What do you mean?

Henryws (changes requested):
image

misc/scripts/search.sh
```diff
@@ -73,8 +90,16 @@ fi
 PACKAGELIST=()
 URLLIST=()
 while IFS= read -r URL; do
+	if [[ ${URL} == "/"* ]] || [[ ${URL} == "~"* ]] || [[ ${URL} == "."* ]]; then
+		sed -i "s#${URL}#file://$(readlink -f ${URL})#g" "$STGDIR/repo/pacstallrepo.txt" ||
+		fancy_message warn "Add `file://` to the local repo absolute path on $CYAN$STGDIR/repo/pacstallrepo.txt$NC`
```
Henryws, line 95: This has a dangling quote "

Henryws (changes requested):

misc/scripts/search.sh
D-Brox: Muscle memory strikes again

Henryws (changes requested):
image
The showed text for local repos should be something like `/home/henry/local-repo-pacstall-test`, instead of having the `file://` in front of it

Henryws (changes requested):
image
This `@` should not be the same color as the repo

misc/scripts/download.sh
```diff
@@ -35,12 +35,14 @@ if check_url "${URL}"; then
 	fi
 
 	case "$URL" in
-		*.pacscript)
-			if ! wget -q --show-progress --progress=bar:force -- "$URL" > /dev/null 2>&1; then
+		*.pacscript|*packagelist)
+			FILE=$(basename "$URL")
+			if ! curl -s -- "$URL" > $FILE; then
```
Henryws, line 40:
```suggestion
			if ! curl -s -- "$URL" > "$FILE"; then
```

Henryws (changes requested):

Henryws (changes requested):
image
Yet the repo does exist. This error also happens with URLs as well

misc/scripts/search.sh
```diff
@@ -28,21 +28,55 @@ if [[ -n $UPGRADE ]]; then
...
+		local REPODIR="$(getPath ${REPO})"
+		echo "\e]8;;$REPO\a$REPODIR\e]8;;\a"
+	else
+		echo "\e]8;;$REPO\a$REPO\e]8;;\a"
+	fi
+}
+
 if [[ $PACKAGE == *@* ]]; then
 	REPONAME=${PACKAGE#*@}
+	if [[ $REPONAME == "file://" ]] ||[[ $REPONAME == "/"* ]] || [[ $REPONAME == "~"* ]] || [[ $REPONAME == "."* ]]; then
```
Henryws, line 77:
```suggestion
	if [[ $REPONAME == "file://" ]] || [[ $REPONAME == "/"* ]] || [[ $REPONAME == "~"* ]] || [[ $REPONAME == "."* ]]; then
```

Henryws (changes requested):

Henryws (approved):

misc/scripts/add-repo.sh
```diff
@@ -49,7 +53,7 @@ else
 	fi
 fi
 
-if ! wget -q --spider -- "$REPO/packagelist"; then
+if ! curl --head -s --fail -- "$REPO/packagelist" > /dev/null; then
```
wizard-28, line 56:
```suggestion
if ! curl --head --location -s --fail -- "$REPO/packagelist" > /dev/null; then
```

Makes it follow redirects.

misc/scripts/download.sh
```diff
@@ -35,12 +35,14 @@ if check_url "${URL}"; then
 	fi
 
 	case "$URL" in
-		*.pacscript)
-			if ! wget -q --show-progress --progress=bar:force -- "$URL" > /dev/null 2>&1; then
+		*.pacscript | *packagelist)
+			FILE=$(basename "$URL")
+			if ! curl -s -- "$URL" > "$FILE"; then
```
wizard-28, line 40:
```suggestion
			if ! curl --location -s -- "$URL" > "$FILE"; then
```

wizard-28 (changes requested):

pacstall
```diff
@@ -194,25 +194,35 @@ function ask() {
...
 			true
-			;;
-		*)
-			fancy_message error "Failed with http code ${http_code}"
-			return 1
-			;;
-	esac
+		fi
+	else
+		http_code="$(curl -o /dev/null -s --head --write-out '%{http_code}\n' -- "${1}")"
```
wizard-28, line 202:
```suggestion
		http_code="$(curl --location -o /dev/null -s --head --write-out '%{http_code}\n' -- "${1}")"
```

Shouldn't this have it too?

wizard-28 (changes requested):

wizard-28 (changes requested):
If the repository is deleted without prior removal from the repolist, then it should warn the user about it, in all commands related to the package repos

It doesn't do that currently

wizard-28 (approved):

PR #543, opened 2022-03-30
fix: use `set -euo pipefail` for pacscript functions

D-Brox:
## Purpose

Errors in pacscript functions are currently ignored unless explicitly caught with a return statement.

## Approach

Put the pacscript functions inside subshells with `set -euo pipefail`.

## Addendum

Closes #537

## Checklist

- [x] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

Issue #537, opened 2022-03-29
Pacscripts should run with 'set -e'


raxod502:
### Is your feature related to a problem? Please describe.

Currently, all errors in Pacscripts are ignored unless explicitly caught with a `return` statement. This default is almost never the correct behavior, and has the potential to mask packaging errors, such as files being installed into nonexistent directories, trying to install files that do not exist, or any compilation errors. This makes it impractical to test Pacscript upgrades in a noninteractive way, since the output of Pacstall must be manually checked to see if any errors occurred that were logged but did not abort execution.

### A clear and concise description of what you want to happen.

By default, all Pacscript hooks should execute in a `set -e` environment. In fact, I would advocate for `set -euo pipefail`, since most programmers generally expect referring to a nonexistent variable to cause an error, and since the non-pipefail behavior (i.e. ignore all errors that may occur in commands on the left-hand side of a pipeline) is a repeated surprise to programmers new to Bash.

This will obviously be a breaking change and will require a number of Pacscripts to be updated, but I think it is an important enough change to justify the inconvenience. To ease the migration, we could make the new default an opt-in to begin with, then migrate over core Pacscripts one at a time, and finally flip the default as a major version bump.

### Describe alternatives you've considered

The main alternative to `set -e` is suffixing every line of every shell script with `|| return`. Some people advocate this approach since it is more explicit, but I tend to think it is at odds with the goal of Pacscript to deduplicate command packaging code and "do the right thing" by default.

### Additional context

`makepkg` uses `set -e` to execute hooks.

wizard-28:
I think since this will entail a possible major refactor of pacscripts, this should be scheduled for the python rewrite.

wizard-28:
Also possibly it would be better to move this issue to LibPacstall, as we're planning to do all the subprocess and other "dirty" stuff in that library, and keeping the clients free of subprocess code as much as we can.

I need @Henryws and @D-Brox's opinion on moving this issue there.

D-Brox:
Works in a subshell, and it doesn't seem like a breaking change given the current codebase.
![image](https://user-images.githubusercontent.com/65723952/160724300-e5c933b7-1a90-4470-a9fa-f5b4d470239e.png)

D-Brox:
I looked better into it, seems like a 4 line change.

Also, @raxod502 can you give me an example of a pacscript that would break with this change?

raxod502:
[...]

PR #608, opened 2022-06-22
fix(install-local): Replace `-ce` with `-ceuo pipefail`

D-Brox:
## Purpose

The original `set -euo pipefail` PR (#543) implemented only `set -e`.

## Approach

Add missing flags when calling bash subprocess

## Addendum

#537

## Checklist


D-Brox:
That's an easy fix on the pacscript tho, just
```bash
randomvar=""
{...}
if [[ -z $randomvar ]]; then
	echo "Random var not set, doing something"
fi
```

Henryws:
As discussed on discord, we're keeping the `-u` flag

Henryws (approved):

wizard-28 (approved):

PR #631, opened 2022-08-14
feat(optdepends): add array selection

Henryws:
## Purpose

Currently, `optdepends` can either be entirely selected or entirely deselected, leaving the user with no choice of what optional dependencies they want to install.

## Approach

Create a prompt that will ask the user which `optdepends` they want to install, or choosing ranges (`2-5`), while also keeping `Y` and `N` options as to select everything or nothing.

## Checklist

##### Using `-P`
image
##### Selecting deps
image
##### None
image
##### Input as non-numbers
image
##### Ranges
image

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -59,7 +59,12 @@ function select_options() {
 	rm -f /tmp/pacstall-select-options
 	local message="${1}"
 	local length="${2}"
-	echo -ne "${message} [${BOLD}$( seq -s ' ' 1 "$length" )${NC}] [${BIGreen}Y${NC}/${RED}n${NC}] "
+
+	if [[ $length -ge 6 ]]; then
+		echo -ne "${message} [${BOLD}1..$length${NC}] [${BIGreen}Y${NC}/${RED}n${NC}] "
+	else
+		echo -ne "${message} [${BOLD}$( seq -s ' ' 1 "$length" )${NC}] [${BIGreen}Y${NC}/${RED}n${NC}] "
+	fi
 	if [[ -z $DISABLE_PROMPTS ]]; then
 		read -r input <&0
 		if [[ $NON_INTERACTIVE ]]; then
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -679,7 +679,7 @@ export srcdir="$PWD"
 sudo chown -R "$PACSTALL_USER":"$PACSTALL_USER" . 2> /dev/null
 
 export pkgdir="/usr/src/pacstall/$name"
-export -f ask fancy_message
+export -f ask fancy_message select_options
 
 # Trap so that we can clean up (hopefully without messing up anything)
 trap cleanup ERR
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -59,7 +59,7 @@ function select_options() {
 	rm -f /tmp/pacstall-select-options
 	local message="${1}"
 	local length="${2}"
-	echo -ne "${message} [$( seq -s ' ' 1 "$length" )] [${BIGreen}Y${NC}/${RED}n${NC}] "
+	echo -ne "${message} [${BOLD}$( seq -s ' ' 1 "$length" )${NC}] [${BIGreen}Y${NC}/${RED}n${NC}] "
 	if [[ -z $DISABLE_PROMPTS ]]; then
 		read -r input <&0
 		if [[ $NON_INTERACTIVE ]]; then
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -77,7 +77,16 @@ function select_options() {
 	elif [[ $input =~ ^[Nn]$ ]]; then
 		echo "n" | tee /tmp/pacstall-select-options >/dev/null
 	elif ! [[ $input =~ [a-zA-Z]+ ]] || [[ $input =~ ^[0-9]+$ ]]; then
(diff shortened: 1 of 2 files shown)


Henryws:
I need to add out of bounds checkers.

wizard-28 (changes requested):
image

If the total number of `optdepends` are more than a handful, then the `[1 2 3]` prompt will become very long. I suggest truncating the series like `[1 .. 20]`

Henryws:
I couldn't find a package that has more than 6 packages, but I modified the neofetch package and duplicated a line
image

wizard-28 (changes requested):
Document the function in the codebase.

wizard-28 (approved):

D-Brox (changes requested):
Move `select_options()` to the main script so it can also be used on `upgrade`'s package selection and for repo selection on `search`

D-Brox (changes requested):
Allow other formatting
image
image

Henryws:
ok so first, `1..5` should be replaced with `1-5`, and second, you want comma separated delimiters as well?

D-Brox:
> ok so first, `1..5` should be replaced with `1-5`, and second, you want comma separated delimiters as well?

@Henryws It's better to have something more robust to user errors

Henryws:
I'll make `1..5` internally turn into `1-5` then.

misc/scripts/install-local.sh
```diff
@@ -198,10 +200,15 @@ function makeVirtualDeb {
 			unset z
 			# tab over the next line
 			echo -ne "\t"
-			ask "Do you want to install the optional dependencies" Y
-			if [[ $answer -eq 1 ]]; then
-				for optdep in "${optdeps[@]}"; do
-					deps+=" ${optdep%%: *}"
+			select_options "Select optional dependencies to install" "${#optdeps[@]}"
+			choices=( $(cat /tmp/pacstall-select-options) )
```
hwittenborn, line 204: Is the output newline separated in the cat command? If so I'd use `mapfile` like that's bot's saying

misc/scripts/install-local.sh
```diff
@@ -198,10 +200,15 @@ function makeVirtualDeb {
...
+			if [[ "${choices[0]}" != "n" ]]; then
+				for i in "${choices[@]}"; do
+					(( i-- ))
+					s="${optdeps[$i]}"
+					if [[ -n $s ]]; then
```
hwittenborn, line 209: Could this just use `"${#s}"` or something so the variable can be quoted? This probably ain't gonna cause issues regardless, but I think Bash best practices should still be followed.

pacstall
```diff
@@ -238,6 +239,61 @@ function download() {
...
+			if [[ -z $input ]]; then
+				echo "Y"
+			fi
+			echo "$input"
+		fi
+	else
+		echo "Y"
+		input="Y"
+	fi
+	if [[ -z $input ]] || [[ $input =~ ^[Yy]$ ]]; then
```
hwittenborn, line 267: Same quoting stuff as the above comment in this function.

pacstall
```diff
@@ -238,6 +239,61 @@ function download() {
...
+						split_arr=(${i//-/ })
+						;;
+					*..*)
+						split_arr=(${i//../ })
+						;;
+					*)
+						select_options "$message" "$length"
+						;;
+				esac
+				out+=( $(seq -s ' ' "${split_arr[0]}" "${split_arr[1]}") )
```
hwittenborn, line 285: This should use `mapfile` too, right?

hwittenborn (changes requested):

misc/scripts/install-local.sh
Henryws: It is space separated.

misc/scripts/install-local.sh
Henryws: I believe variables inside `[[` don't need to be quoted.

pacstall
Henryws: This command will have predictable output so I don't think it's needed.

misc/scripts/install-local.sh
hwittenborn: Fair enough

misc/scripts/install-local.sh
hwittenborn: They probably don't *need* to, but I still think it'd be best if they were still

pacstall
hwittenborn: I still think the mapfile approach would be cleaner looking though:

```sh
mapfile -O "${#out[@]}" -t out < <(seq -s ' ' "${split_arr[0]}" "${split_arr[1]}")
```

Henryws: > I still think the mapfile approach would be cleaner looking though:
> 
> ```sh
> mapfile -O "${#out[@]}" -t out < <(seq -s ' ' "${split_arr[0]}" "${split_arr[1]}")
> ```

That doesn't seem to look "cleaner" tbh

misc/scripts/install-local.sh
Henryws: Our policy at pacstall is don't quote variables inside `[[ ]]` unless they somehow need to.

hwittenborn: Fair enough

D-Brox (approved):

pacstall
hwittenborn: It's just avoiding nested `()` stuff I guess, it won't bother me too much if it's not changed

PR #632, opened 2022-08-25
feat(upgrade.sh): Parallel package version comparison on upgrade

D-Brox:
## Purpose

The upgrade is too slow, and this gets worse when there is more packages in the repos and packages installed

## Approach

Multi-processing on the version fetch loop, reducing the bottleneck.

## Checklist

### DIFF
--- a/misc/scripts/upgrade.sh
+++ b/misc/scripts/upgrade.sh
@@ -43,7 +43,7 @@ touch /tmp/pacstall-up-urls
 
 fancy_message info "Checking for updates"
 
-N=$(nproc)
+N="$(nproc)"
 (
 for i in "${list[@]}"; do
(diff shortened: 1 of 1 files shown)


misc/scripts/upgrade.sh
```diff
@@ -43,7 +43,10 @@ touch /tmp/pacstall-up-urls
 
 fancy_message info "Checking for updates"
 
+N=$(nproc)
```
Henryws, line 46:
```suggestion
N="$(nproc)"
```

Henryws (changes requested):

Henryws (approved):

wizard-28 (approved):

PR #654, opened 2022-10-09
fix: cat pacscript into post files

Henryws:
## Purpose

Some packages require it.

## Checklist


D-Brox:
> Some packages require it
This bug was found when it nuked my `/usr/share/`

wizard-28 (approved):

PR #655, opened 2022-10-09
fix: all packages can use `gives` now

Henryws:
## Purpose

Currently, only debs can make use of `gives`. This PR makes it so that any package can be installed under a name besides whats declared in `$name`.

## Checklist


wizard-28 (changes requested):
Add provides array

wizard-28 (approved):
LGTM

PR #666, opened 2022-10-12
fix(prompt_optdepends): add guard to max optdepend

Henryws:
## Purpose

Currently inputing invalid numbers into `prompt_optdepends` will just be skipped over entirely instead of giving a warning.

## Checklist


httpllamaz (approved):
LGTM

PR #701, opened 2022-10-29
fix(fish): Build completions

D-Brox:
## Purpose

Fish completions were broken with the addition of the `-B` flag.

## Approach

Try not to lose my mind fixing it

## Progress

It's usable now

## Checklist


Henryws (changes requested):
When running `pacstall `, then tabbing, no flags appear. I must add a dash myself before they appear.

Henryws (approved):

PR #717, opened 2022-11-11
feat: add `incompatible` array

### LINKED ISSUE #698: support distribution differences through metadata or conditional evaluation
# Problem

The Wiki provides the following explanation:

> Pacstall is intended for use on Ubuntu based distributions, however most packages in our repository work on Debian. A couple Pacscripts may not work as they might have outdated dependencies or a dependency name change, but for the most part, Pacstall does work fine on Debian.

That is, build scripts are targeted for Ubuntu systems, and it remains a matter for chance whether any may work under  vanilla Debian.

# Request

Two successive stages of development may be considered to provide better support multiple distributions, as follows:

1. Declarations in the build script to designate the distributions for which a script may be considered explicitly to support (or explicitly not to support).
2. Conditional evaluation of script values, to provide support for multiple distributions, despite differences, in the same build script.

Note that for some purposes   releases, of a distribution, not just the overall project, are useful distinctions.

Also note that some derived distributions, such as Linux Mint, are designated by an Ubuntu base release as well as an LM release.

# Notes

The proposal given appears to be the most general to support all Debian-based, or APT-based, operating systems,  through a unified build system, without regressing current functionality or level of support.

Henryws:
## Purpose

Currently some scripts have methods of preventing running on certain distros using this codeblock:
```bash
if [[ "$(grep -Po 'ID=\K[^"]+' /etc/os-release)" == "debian" ]]; then
    fancy_message error "This script doesn't work on Debian"
    cleanup # Trigger pacstall cleanups
    exit 1
fi
```

This is not the best as it triggers functions that should really only be called in pacstall, it clutters the script, it doesn't take into account version differences, etc. It also prevents some CIs from sourcing this at all.

## Approach

Add a built in method with a new array called `incompatible`. Each key should look like this: `$distro:$version_name`. This will declutter a lot of scripts, and make it easier to fine tune Pacscripts. Globs also work. This will mean that a Pacscript has no running components when sourced (besides `pkgver` in some).

## Addendum

Closes #698 

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -154,6 +154,18 @@ function compare_remote_version() (
     fi
 )
 
+function get_incompatible_releases() {
+    local distro_name="$(lsb_release -si 2> /dev/null | tr '[:upper:]' '[:lower:]')"
+    local distro_version_name="$(lsb_release -sc 2> /dev/null)"
+    local distro_version_number="$(lsb_release -sr 2> /dev/null)"
+    local input=("$@")
+    for key in "${input[@]}"; do
+        if [[ $key == "${distro_name}:${distro_version_name}" ]] || [[ $key == "${distro_name}:${distro_version_number}" ]]; then
+            return 1
+        fi
+    done
+}
+
 function deblog() {
     local key="$1"
     local content="$2"
@@ -188,7 +200,7 @@ function prompt_optdepends() {
             fi
             # Add to the dependency list if already installed so it doesn't get autoremoved on upgrade
             # If the package is not installed already, add it to the list. It's much easier for a user to choose from a list of uninstalled packages than every single one regardless of it's status
-			if ! [[ "$(dpkg-query -W -f='${Status}' "${opt}" 2> /dev/null)" == "install ok installed" ]]; then
+            if ! [[ "$(dpkg-query -W -f='${Status}' "${opt}" 2> /dev/null)" == "install ok installed" ]]; then
                 optdeps+=("${optdep}")
             else
                 deps+=("${opt}")
@@ -319,11 +331,11 @@ function createdeb() {
 }
 
 function makedeb() {
-	if [[ -n $gives ]]; then
+    if [[ -n $gives ]]; then
         fancy_message info "Packaging ${BGreen}$name${NC} as ${BBlue}$gives${NC}"
-	else
+    else
         fancy_message info "Packaging ${BGreen}$name${NC}"
-	fi
+    fi
     deblog "Package" "${gives:-$name}"
 
     if [[ $version =~ ^[0-9] ]]; then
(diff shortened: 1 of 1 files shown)


brainchild0:
May I ask, why do you choose to support the  specification of  *incompatible* releases, rather than *compatible* ones?

Suppose some contributor creates a script that functions on some releases, but not others, and then later another contributor wishes to add a script that functions on some of the releases not supported by the other? 

Would it not be preferred to adopt a method that allows the tool to select a script appropriate for the local system, even if more than one is defined for some package name?

Henryws:
Because incompatible implicitly allows everything except certain systems, rather than specifying literally every distro and version possible

brainchild0:
Yes, if no value is given, then all distributions should match. '

However, if a criteria is given for distinguishing releases, then it may be preferred that the criteria should be inclusive rather than exclusive. In fact, if the maintainer has tested on certain systems, then it would be difficult to identify such systems by excluding all other possibilities.

misc/scripts/install-local.sh
```diff
@@ -154,6 +154,22 @@ function compare_remote_version() (
     fi
 )
 
+function get_incompatible_releases() {
+    local distro_name="$(lsb_release -si 2> /dev/null | tr '[:upper:]' '[:lower:]')"
```
D-Brox, line 158: This doesn't take into account what distro your system is based on. So if `incompatible=("ubuntu:jammy")`, it would install on Pop_os jammy

misc/scripts/install-local.sh
```diff
@@ -154,6 +154,22 @@ function compare_remote_version() (
...
+    if [[ "$(lsb_release -ds 2> /dev/null | tail -c 4)" == "sid" ]]; then
+        local distro_version_name="sid"
+    else
+        local distro_version_name="$(lsb_release -sc 2> /dev/null)"
+    fi
+    local distro_version_number="$(lsb_release -sr 2> /dev/null)"
+    local input=("$@")
+    for key in "${input[@]}"; do
+        if [[ $key == "${distro_name}:${distro_version_name}" ]] || [[ $key == "${distro_name}:${distro_version_number}" ]]; then
```
D-Brox, line 167: `key` should be normalized to lowercase, that way both `Ubuntu:Jammy` and `ubuntu:jammy` would work as key

misc/scripts/install-local.sh
D-Brox: Accepting `"*:jammy"` as key would also work

D-Brox (changes requested):

misc/scripts/install-local.sh
brainchild0: Perhaps it be more helpful if the (release) codename and the OS ID (distribution)   were given as completely separate fields, rather than as a single compound field.

Henryws: > This doesn't take into account what distro your system is based on. So if `incompatible=("ubuntu:jammy")`, it would install on Pop_os jammy

Any idea how to fix this?

brainchild0: I think the suggestion was a way to specify a particular release codename, but still match all distributions.

Henryws: You can now do something like `incompatible=("*:jammy")`

D-Brox (approved):

wizard-28 (approved):

PR #731, opened 2022-12-12
fix(depends): install depends before build

### LINKED ISSUE #727: bug: Runtime dependencies are not installed before building
### Describe the bug

Meson packages check the runtime packages before compiling the package. But pacstall doesn't install the packages before compiling it

### To reproduce

Install this meson package:

```bash
name="g4music"
gives="g4music"
version="1.8.1"
build_depends="meson valac"
depends="libgtk-4-dev libgstreamer1.0-dev libadwaita-1-dev"
url="https://gitlab.gnome.org/neithern/g4music/-/archive/v${version}/g4music-v${version}.tar.gz"
repology=("project: g4music")
description="Play your music in an elegant way"
maintainer="সৌম্যদীপ <soumyadeepghosh2004@zohomail.in>"

prepare() {
  true
}

build(){
	meson setup build --buildtype=release
}

install(){
	sudo meson install -C build --destdir=${pkgdir}
}
```

### Expected behavior

All the deps should have been installed properly

### Screenshots

NA

### OS version

Ubuntu 22.10

### Pacstall version

3.0.2

Henryws:
## Addendum

Fixes #727.

## Checklist


D-Brox (commented):
I don't see how this "fix" is relevant. Apt installs the runtime deps along with the package. If the a runtime dep is needed during build time, then it should be in both dep arrays.

wizard-28:
> I don't see how this "fix" is relevant. Apt installs the runtime deps along with the package, as it should and it's supposed to. If the a runtime dep is needed during build time, then it should be in both dep arrays.

If they're in both arrays, wouldn't Pacstall offer to uninstall them after the package is installed? That might break the package.

D-Brox:
Then that's what should be fixed. It should be done in a way similar to how `mk-build-deps` works, modified to our need:
- Explicitly install the build deps if you want to keep.
- Use a dummy package with all build deps as dependencies if you want to remove.
That way we can safely uninstall the dummy and autoremove the packages.

Henryws:
What if I find common keys within both `depends` and `build_depends`, and those that exist will be not considered when `build_depends` is removed. Then I could set that package to `auto` installed after the deb so that it will be removed after the deb is uninstalled.

soumyatheman:
> What if I find common keys within both `depends` and `build_depends`, and those that exist will be not considered when `build_depends` is removed. Then I could set that package to `auto` installed after the deb so that it will be removed after the deb is uninstalled.

That's I guess the best approach.




> I don't see how this "fix" is relevant. Apt installs the runtime deps along with the package, as it should and it's supposed to. If the a runtime dep is needed during build time, then it should be in both dep arrays.

No, as meson first assures, that after building the app, it'll run properly. This problem is specially with meson builds. But the the deps, if tagged as build deps, which in reality not, will be messed. **Build Deps** are hardly ever **Runtime Deps**

PR #753, opened 2022-12-29
fix(remove): remove instead of purge

Henryws:
## Purpose

We don't need to purge.

## Approach

Remove instead of purging.

## Checklist


wizard-28:
Why

Henryws:
Because purging also runs removescript twicem

wizard-28 (approved):

PR #760, opened 2023-01-04
fix(ppa): add dangling PPA warning

Henryws:
## Purpose

PPAs linger after a package is removed that depended on a PPA.

## Approach

Remove them in `remove.sh` and convert the `_ppa` string into an array.

## Checklist


wizard-28 (changes requested):
This would also remove PPAs that the user has installed manually, but also happens to be used in a pacscript.

Henryws:
How about we print a warning telling the user a potential dangling PPA is still on the system and provide a command to run to remove it if needed?

wizard-28 (approved):

PR #767, opened 2023-01-07
fix(prompt_optdepends): prevent only adding non-installed pkgs to suggests

Henryws:
## Purpose

Move `optdepends` to `Suggests` and install them afterwards.
Change some variable names in order to make `prompt_optdepends` readable.
Remove `prompt_optdepends` entirely when using `-B` flag.
Install optdepends before package builds (now can specify buildtime optdepends)

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -499,7 +499,15 @@ Pin: version *
 Pin-Priority: -1" | sudo tee /etc/apt/preferences.d/"${name}-pin" > /dev/null
         if [[ -n ${not_installed_yet_optdeps[*]} ]]; then
             fancy_message info "Installing selected optional dependencies"
+            for pkg in "${not_installed_yet_optdeps[@]}"; do
+                if [[ "$(dpkg-query -W -f='${Status}' "${pkg}" 2> /dev/null)" != "install ok installed" ]]; then
+                    local to_mark_as_auto+=("$pkg")
+                fi
+            done
             sudo -E apt-get install ${not_installed_yet_optdeps[*]} -y 2> /dev/null
+            if [[ -n $to_mark_as_auto ]]; then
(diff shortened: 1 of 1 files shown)


wizard-28:
> Move optdepends to Suggests and install them afterwards.

This doesn't allow for optional features when building then.

Henryws:
> > Move optdepends to Suggests and install them afterwards.
> 
> This doesn't allow for optional features when building then.

Should we make `build_optdepends` then?

wizard-28:
> > > Move optdepends to Suggests and install them afterwards.
> > 
> > 
> > This doesn't allow for optional features when building then.
> 
> Should we make `build_optdepends` then?

What's the rationale behind moving `optdepends` to `Suggests` in the first place?

Henryws:
Because `Suggests` is where optional dependencies are supposed to actually go in the first place according to Debian.

wizard-28:
> Because `Suggests` is where optional dependencies are supposed to actually go in the first place according to Debian.

Which pacscripts that use `optdepends` currently, can switch to the new `optdepends` without breaking?

Henryws:
Unless they use `optdepends` for optional building programs (which I don't think any do), it will break, but if it's purely used for enhancing the finished binary/program/whatever (which is the entire point of `optdepends`, it's fine.

wizard-28:
>  but if it's purely used for enhancing the finished binary/program/whatever (which is the entire point of optdepends, it's fine.

Give examples of those.

Henryws:
https://github.com/pacstall/pacstall-programs/blob/bc5ebe0228fee9f4e05a02755855eb153f30017e/packages/os-installer-git/os-installer-git.pacscript

wizard-28:
I don't think it's worth introducing a breaking change for this.

wizard-28 (approved):

PR #769, opened 2023-01-11
fix(homedir): detect any home dir, not just /home

Henryws:
## Purpose

Currently we set `homedir` to `/home/$USER`, but this can be an issue when users *could* in theory set their homedir to something else.

## Tests
```bash
henry@plasma:~$ echo ~root
/root
henry@plasma:~$ echo ~henry
/home/henry
henry@plasma:~$ echo ~games
/usr/games
henry@plasma:~$ echo ~irc
/run/ircd
```
vs setting the assumed homedir to `/home/$USER`.

## Drawbacks
This approach has one drawback and that is that is forces the Deb's post{rm,inst} running shell to be bash, not sh, however this isn't such a big deal considering that you have to have bash installed to run Pacstall, and most systems that a `-B` Deb is being installed on should have it as well.

## Approach

Use the `~$USER` bashism to get the canonical path to any given usernames homedir.

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -449,7 +449,7 @@ function fancy_message() {
 
 function get_homedir() {
 	local PACSTALL_USER=$(logname 2> /dev/null || echo "${SUDO_USER:-${USER}}")
-	echo "/home/$PACSTALL_USER"
+	echo ~"$PACSTALL_USER"
 }
 export homedir="$(get_homedir)"
 
@@ -550,7 +550,7 @@ fi
 
 fancy_message info "Sourcing pacscript"
 DIR=$(pwd)
-homedir="/home/$PACSTALL_USER"
+homedir=~"$PACSTALL_USER"
 export homedir
 
 pacfile=$(readlink -f "$PACKAGE".pacscript)


Henryws:
What do you mean?

wizard-28:
> What do you mean?

What happens if `/home/irc` exists?

wizard-28 (approved):

Qussayyon:
@Henryws 
Does this work with user `root` ?

Henryws:
Run `echo ~root` in bash

Henryws:
Pacstall isn't intended to run as root. This PR was made because some people *might* have a non-standard home directory.

Qussayyon:
Great, got it.
Another question:
In my computer: when I run:  `echo ~"$PACSTALL_USER"`
I get only `~rwd` but not `/home/rwd`
Any idea why?

Henryws:
What shell are you using?

Qussayyon:
bash

```bash
echo $BASH_VERSION
5.1.16(1)-release
```

Qussayyon:
This is what I run so far:
image

Qussayyon:
True.

image

Henryws:
Issue was that I made this PR testing in ZSH, where it works without evaling 😊

Henryws:
I released #776 @Qussayyon. The issue should be fixed.

Qussayyon:
True, I first tested it in `zsh` without noticing that you base your code because of `bash` but when I used it in `bash` it showed me the issue.

I'm glad I helped.

PR #770, opened 2023-01-13
ref(pwd): `pwd` -> `$PWD`

Henryws:
## Purpose

It serves no purpose to call an external command when a builtin bash variable does the exact same thing.

## Checklist


wizard-28:
> It serves no purpose to call an external command when a builtin bash variable does the exact same thing.

`pwd` is a bash builtin though.

```
~
✦ ❯ type pwd
pwd is a shell builtin
```

Henryws:
`which pwd`

PR #790, opened 2023-02-19
fix(cleanup): unset all variables during life of pacscript

Henryws:
## Purpose

If some variables are not cleaned out, they can leak into others, which happened after updating steamtinkerlaunch-git, then rhino-pkg-git. The `postinst` was leaked into rhino-pkg-git when it had none, resulting in an install error.

## Approach

I went through the Pacscript 101 wiki and made sure everything was cleaned.

## Checklist


D-Brox (approved):

volitank (approved):
LGTM

PR #808, opened 2023-02-28
ref(search): replace grep+cut+awk with awk

D-Brox:
## Purpose

Remove `grep` from `search.sh`

## Approach

Use `awk` instead

## Checklist

- [X] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

### DIFF
--- a/misc/scripts/search.sh
+++ b/misc/scripts/search.sh
@@ -84,7 +84,7 @@ if [[ $PACKAGE == *@* ]]; then
         specifyRepo "$URL"
         if [[ $URLNAME == "$REPONAME" ]]; then
             mapfile -t PACKAGELIST < <(curl -s -- "$URL"/packagelist)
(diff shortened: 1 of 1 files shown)


Henryws:
Replace the same code in `upgrade.sh`, line 72.

D-Brox:
> ~~Replace the same code in `upgrade.sh`, line 72.~~
> Actually it looks different, so I'll check it out.
> image
> 👀 

Did you resize the terminal or did it print that way?

Henryws:
It printed that way because we run stuff with `&` during the check, and there's *supposed* to be no output, so when it does it appears all funky.

Henryws (approved):

PR #809, opened 2023-02-28
fix(download): always create `SRCDIR`

Henryws:
## Purpose

`SRCDIR` doesn't get remade after a package upgrades, meaning the download will fail on upgrades.

## Checklist

### DIFF
--- a/misc/scripts/download.sh
+++ b/misc/scripts/download.sh
@@ -25,17 +25,15 @@
 # This script downloads pacscripts from the interwebs
 
 if check_url "${URL}"; then
-    if [[ $type == "install" ]]; then
-        mkdir -p "$SRCDIR" || {
-            if ! [[ -w $SRCDIR ]]; then
-                suggested_solution "Run '${UCyan}sudo chown -R $PACSTALL_USER:$PACSTALL_USER $SRCDIR'${NC}"
-            fi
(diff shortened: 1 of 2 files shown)


D-Brox (changes requested):
Don't patch `download.sh`, patch `upgrade.sh`.

The change should be moving this and the following lines into the for loop:

https://github.com/pacstall/pacstall/blob/7f90fa8daac33943c15d5aa3a8f41ec6a0555bf4/misc/scripts/upgrade.sh#L142

misc/scripts/upgrade.sh
```diff
@@ -139,8 +139,6 @@ ${BOLD}$(cat /tmp/pacstall-up-print)${NORMAL}\n"
     done < /tmp/pacstall-up-urls
 
     export local='no'
-    sudo mkdir -p "$SRCDIR"
-    sudo chown -R "$PACSTALL_USER":"$PACSTALL_USER" "$SRCDIR"
     if ! cd "$SRCDIR" 2> /dev/null; then
```
D-Brox, line 142: The `if` should also go into the for loop

D-Brox (changes requested):

D-Brox (approved):

PR #834, opened 2023-03-13
fix(pacdeps): move pacfile to `/tmp`

Henryws:
## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -45,6 +45,7 @@ function cleanup() {
     unset name pkgname repology epoch url depends build_depends breaks replace gives description hash optdepends ppa arch maintainer pacdeps patch PACPATCH NOBUILDDEP provides incompatible optinstall epoch pac_functions 2> /dev/null
     unset -f pkgver postinst removescript prepare build install 2> /dev/null
     sudo rm -rf "${SRCDIR:?}"/*
+    sudo rm "${pacfile:?}"
 }
 
 function trap_ctrlc() {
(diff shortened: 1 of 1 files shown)


misc/scripts/install-local.sh
```diff
@@ -571,9 +572,10 @@ DIR="$PWD"
 homedir="$(eval echo ~"$PACSTALL_USER")"
 export homedir
 
-pacfile=$(readlink -f "$PACKAGE".pacscript)
+sudo mv "${PACKAGE}.pacscript" /tmp
```
D-Brox, line 575: Use `cp` instead. We don't want to move an user-made pacscript out of their workspace.

D-Brox (changes requested):
Add the temporary `pacfile` to cleanup

pacstall
```diff
@@ -596,12 +596,13 @@ Helpful links:
...
                     fancy_message error "Failed to install ${GREEN}${PACKAGE}${NC}"
-            		rm -rf "${SRCDIR:?}"
+                    if ! [[ -f "/tmp/pacstall-pacdeps-$PACKAGE" ]]; then
+                        rm -rf "${SRCDIR:?}"
+                    fi
                     exit 1
                 fi
                 shift
             done
-            rm -rf "${SRCDIR:?}"
```
D-Brox, line 604: Keep this, but with the same `pacdep` check

D-Brox (changes requested):

pacstall
D-Brox: Or else `/tmp/pacstall` doesn't get deleted

Henryws: I tried with:
```bash
if ! [[ -f "/tmp/pacstall-pacdeps-$PACKAGE" ]]; then
	sudo rm -rf "${SRCDIR:?}"
fi
```
and it didn't work..

D-Brox (dismissed):

D-Brox (approved):

D-Brox (approved):

wizard-28 (approved):

PR #839, opened 2023-03-16
fix(pacdeps): pacdeps are implicit deps to parent package

Henryws:
## Purpose

Previously we'd check if deps exist first, then if pacdeps exist, then make them implicit deps, but we don't want that. We want to check if pacdeps exist, then add them to the deps list.

## Checklist


oklopfer (approved):
lgtm

wizard-28 (approved):
Force push to master 🚀

PR #841, opened 2023-03-16
refactor: quote variables & use `mapfile` wherever possible

Henryws:
## Purpose

Safety:tm: by using `mapfile` to convert a string at a delimiter instead of relying on bash to do it, and also by quoting variables. 

This also fixes `select_options` to allow for complex stuff like `1..3 5-10` which it should have been able to since the beginning.

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -341,7 +341,7 @@ function select_options() {
         echo -ne "${message} [${BOLD}$(seq -s ' ' 1 "$length")${NC}] [${BIGreen}Y${NC}/${RED}n${NC}] "
     fi
     if [[ $DISABLE_PROMPTS == "no" ]]; then
-        read -r input <&0
+        read -ra input <&0
         if [[ $NON_INTERACTIVE ]]; then
             if [[ -z $input ]]; then
                 echo "Y"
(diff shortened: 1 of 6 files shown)


pacstall
```diff
@@ -634,6 +645,7 @@ Helpful links:
             while [[ -n $2 ]]; do
                 PACKAGE=$2
                 shift
+                # shellcheck source=./misc/scripts/search.sh
```
wizard-28, line 648:
```suggestion
                # shellcheck source=./misc/scripts/remove.sh
```

pacstall
```diff
@@ -750,11 +762,10 @@ Helpful links:
...
                 if ! source "$STGDIR/scripts/search.sh"; then
-                    ((i++))
                     continue
                 fi
 
-                specifyRepo $REPO
+                specifyRepo "$REPO"
                 export URL="$REPO/packages/$PACKAGE/$PACKAGE.pacscript"
 
                 if ! source "$STGDIR/scripts/download.sh"; then
```
wizard-28, line 771: Are the shellcheck comments not required here?

wizard-28 (changes requested):

wizard-28 (approved):

wizard-28 (approved):

wizard-28 (approved):

PR #872, opened 2023-04-09
fix: only delete stuff if not in payload mode

wizard-28:
## Purpose

Fixes payload bug, where it'd delete the file before processing it.

## Approach

Check if we're in payload mode, and proceed accordingly

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -300,7 +300,9 @@ function check_url() {
 
 # use axel if available
 function download() {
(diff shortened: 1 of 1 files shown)


Henryws (changes requested):
Run shfmt on it (`shfmt -bn -ci -sr -s -i 4 -w pacstall`)

Henryws (approved):

PR #881, opened 2023-04-21
fix(PACSTALL_PAYLOAD): skip if pacdep

Elsie19:
## Purpose

If a package is being downloaded and it has a pacdep, `PACSTALL_PAYLOAD` will fail. We want the main package not downloaded, but the pacdeps are fine.

## Checklist


wizard-28 (approved):

D-Brox:
I suggest using a `payload=PACSTALL_PAYLOAD_$PACKAGE` and `${!payload}` to solve the issues with pacdeps, or something along those lines

PR #926, opened 2023-06-15
feat!(pacscript): convert variables to PACBUILD equivs

Elsie19:
## Purpose

Because it'll ease the transition to PACBUILDs.

## Progress

- [x] `description` -> `pkgdesc`
- [x] `version/pkgver()` -> `pkgver/pkgver()`
- [x] `build_depends` -> `makedepends`
- [x] `postinst` -> `post_install`
- [x] `preinst` -> `pre_install`
- [x] `removescript` -> `post_remove`
- [x] `pkgrel`

#### `version/pkgver()` -> `pkgver/pkgver()`
Historically, we have had `version` be the canonical entrypoint to get a version, whether that be through a string, or for git packages, by command substitution to the `pkgver()` function. What will change now is we'll have a variable and function of the same name, called `pkgver`. For non-git packages, the only change that needs to happen is to change `version` -> `pkgver`. For git packages, you must include both the variable and function, with the function being a way to get the version, and the variable being the latest version from `pkgver()` at the time of pushing. Pacstall developers are currently discussing how they should be formatted.

#### `description` -> `pkgdesc`
Simply replace them, no other changes needed.

#### `build_depends` -> `makedepends`
Simply replace them, no other changes needed.

## Other

There are some variables I don't want to transition in any of the bash versions, like `url` -> `source` and consequently `homepage` -> `url`. These changes in this PR should be almost *sed*able. Basically, this PR will make a proto-PACBUILD.

Changed `checks()` to cascade errors instead of failing out at the first error so the developer can see all errors at once.

## Checklist


D-Brox:
Should `update.sh` migrate the `-git` packages to the new versioning? If not, then `upgrade.sh` should take into consideration that the versioning went from the unix epoch/calver hybrid we were using to the correct PACBUILD spec git versioning

Elsie19:
> epoch/calver hybrid
We used the sha1 commit hash.

PR #929, opened 2023-06-21
feat(upgrade): implement dependency tree

Elsie19:
## Purpose

Packages like [rhino-core](https://github.com/pacstall/pacstall-programs/blob/master/packages/rhino-core/rhino-core.pacscript) and [linux-kernel](https://github.com/pacstall/pacstall-programs/blob/master/packages/linux-kernel/linux-kernel.pacscript) are metapackages that hold calls to install other pacscripts; and due to how the pacdeps function works, a specific order to install those pacscripts. If the dependencies of these metapackages are installed before they are called for the proper order, installs may fail. The dependency tree aims to smooth this out.

## Approach

Associative array magic 🪄. Basically we yeet out pacdeps from the update equation, because `-I` checks for updates of pacdeps, then we just order everything.

## Other

> **Note**
> Order printing is not correct, but it updates in the correct order. May or may not work on that.

## Checklist


Henryws:
Currently an issue reported by Oren that if a pacdep is updated, but not the parent, the pacdep won't update.

oklopfer:
> Currently an issue reported by Oren that if a pacdep is updated, but not the parent, the pacdep won't update because it's stripped from the upgrade process.

Clarification: as an example, rhino-pkg-git is a pacdep of rhino-core. With the dep-tree in its current state, `pacstall -Up` will only find if rhino-core updates; if rhino-pkg-git updates separately from the core, it is not found by the `Up` process.

Henryws:
If the pacdep's dependee (parent package) is not in the upgrade, we should upgrade the pacdep, but if the parent package is in the upgrade, disregard it because `-I` takes care of pacdep upgrades.

misc/scripts/upgrade.sh
```diff
@@ -22,23 +22,39 @@
...
 # Get the list of the installed packages
-mapfile -t list < <(pacstall -L)
+mapfile -t list < <(pacstall -L | awk NF)
+if ((${#list[@]} == 0)); then
+    fancy_message info "Nothing to upgrade"
+    return 0
+fi
+fancy_message info "Building dependency tree"
+dep_tree.loop_traits update_order "${list[@]}"
+dep_tree.trim_pacdeps update_order
```
oklopfer, line 51: get rid of him

oklopfer (commented):
good work, soldier. LGTM 🛸

oklopfer (approved):

PR #933, opened 2023-07-02
refactor(date): replace subshell calls to `date` with `printf`

Elsie19:
## Purpose

`$(date)` is way slower than `printf '%()T'`, which is a builtin.

## Approach

Replace them with equivalent calls.

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -125,7 +125,7 @@ function log() {
         echo "_name=\"$name"\"
         echo "_version=\"${full_version}"\"
         echo "_install_size=\"${install_size}"\"
-        echo "_date=\"$(date)"\"
+        printf '_date=\"%(%a %b %_d %r %Z %Y)T\"\n'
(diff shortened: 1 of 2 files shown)


D-Brox (approved):

misc/scripts/install-local.sh
```diff
@@ -125,7 +125,7 @@ function log() {
         echo "_name=\"$name"\"
         echo "_version=\"${full_version}"\"
         echo "_install_size=\"${install_size}"\"
-        echo "_date=\"$(date)"\"
+        printf '_date=\"%(%a %b %_d %r %Z %Y)T\"\n'
```
D-Brox, line 128: Day before month

D-Brox (changes requested):

misc/scripts/install-local.sh
Elsie19: No.

D-Brox (approved):

PR #942, opened 2023-07-11
fix(pacstall): add `FARCH` array

oklopfer:
## Purpose

the is_compatible_arch check doesn't currently work for foreign architectures - i.e. if dpkg adds arm64 to an amd64 machine, CARCH won't do, so we need a solution.

## Approach

Make a FARCH variable array, test if it exists, and test if our input arch matches one of the FARCH outputs

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -241,9 +241,12 @@ function is_compatible_arch() {
     elif [[ " ${input[*]} " =~ " any " ]]; then
         return 0
     elif [[ -n "${FARCH[*]}" ]]; then
-        if [[ " ${input[*]} " =~ " ${FARCH[*]} " ]]; then
+        if [[ " ${FARCH[*]} " =~ " ${input[*]} " ]]; then
             fancy_message warn "This package is for ${BBlue}${input[*]}${NC}, which is a foreign architecture"
             return 0
+        else
+            fancy_message warn "This Pacscript does not work on ${BBlue}${CARCH}${NC}"
+            return 1
         fi
     elif ! [[ " ${input[*]} " =~ " ${CARCH} " ]]; then
         fancy_message error "This Pacscript does not work on ${BBlue}${CARCH}${NC}"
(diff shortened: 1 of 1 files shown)


misc/scripts/install-local.sh
```diff
@@ -240,6 +240,11 @@ function is_compatible_arch() {
         return 0
     elif [[ " ${input[*]} " =~ " any " ]]; then
         return 0
+    elif [[ -n "${FARCH[*]}" ]]; then
+        if [[ " ${input[*]} " =~ " ${FARCH[*]} " ]]; then
+            fancy_message warn "This package is for a ${BBlue}foreign architecture${NC}"
```
Elsie19, line 245: Specify the architecture.

Elsie19 (changes requested):

misc/scripts/install-local.sh
```diff
@@ -241,8 +241,18 @@ function is_compatible_arch() {
...
+            	fancy_message warn "This package is for ${BBlue}${input[*]}${NC}, which is a foreign architecture"
+            	return 0
+			else
+           		fancy_message error "This Pacscript does not work on ${BBlue}${CARCH}${NC}"
+        		return 1
+			fi
+   		else
+        	fancy_message error "This Pacscript does not work on ${BBlue}${CARCH}${NC}"
+        	return 1
+		fi
```
Elsie19, line 255:
```suggestion
        if [[ -n ${FARCH[*]} ]]; then
            if [[ " ${FARCH[*]} " =~ " ${input[*]} " ]]; then
                fancy_message warn "This package is for ${BBlue}${input[*]}${NC}, which is a foreign architecture"
                return 0
            else
                fancy_message error "This Pacscript does not work on ${BBlue}${CARCH}${NC}"
                return 1
            fi
        else
            fancy_message error "This Pacscript does not work on ${BBlue}${CARCH}${NC}"
            return 1
        fi

```

Elsie19 (changes requested):

Elsie19 (approved):

PR #943, opened 2023-07-11
fix(-L): hide no install error on non-interactive

Elsie19:
## Purpose

Because we just want to exit, and not show anything to the program trying to access the package list.

## Approach

Check for `[[ -t 1 ]]`.

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -808,6 +808,8 @@ Helpful links:
             if ((${#packages_installed[@]} == 0)) && [[ -t 1 ]]; then
                 fancy_message error "Nothing installed yet"
                 exit 1
+            else
+                exit 1
             fi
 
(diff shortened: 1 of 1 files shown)


pacstall
```diff
@@ -805,7 +805,7 @@ Helpful links:
             cd "${METADIR}" || exit 1
             shopt -s nullglob
             packages_installed=(*)
-            if ((${#packages_installed[@]} == 0)); then
+            if ((${#packages_installed[@]} == 0)) [[ -t 1 ]]; then
```
Elsie19, line 808:
```suggestion
            if ((${#packages_installed[@]} == 0)) && [[ -t 1 ]]; then
```

PR #962, opened 2023-08-12
refactor(pacdeps): do not print installed message

Elsie19:
## Purpose

It's not needed.

## Checklist


oklopfer (changes requested):
actually this is needed to catch errors

PR #994, opened 2023-10-16
fix: install pacdeps from the same origin

### LINKED ISSUE #680: Installing item with multiple repositories, dependencies defaults to default repo
### Describe the bug

Installing the `linux-kernel` package from https://github.com/rhino-linux/repo has dependencies default to the default Pacstall repository, which is a patch behind. Therefore it's installing the 6.0 Linux kernel instead of 6.0.1

### To reproduce

Add the Rhino Linux repository: `pacstall -A https://raw.githubusercontent.com/rhino-linux/repo/master`

Install linux-kernel `pacstall -I linux-kernel`

Choose to install via rhino-linux/repo

### Expected behavior

Installs dependencies via `rhino-linux` repository

### Screenshots

image

### OS version

Ubuntu 22.04

### Pacstall version

2.0.1 Mellow

D-Brox:
## Purpose

Currently, pacdeps are being installed from the main repo if they exist there instead of giving preference to the same repo as the package being installed

## Approach

Check if the pacdep exists in the same repo as the package that requires it and install from it if it exists

## Addendum

Closes #680

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -49,7 +49,7 @@ function cleanup() {
     fi
     sudo rm -rf "${STOWDIR}/${name:-$PACKAGE}.deb"
     rm -f /tmp/pacstall-select-options
-    unset name repology pkgver epoch url depends makedepends breaks replace gives pkgdesc hash optdepends ppa arch maintainer pacdeps patch PACPATCH NOBUILDDEP provides incompatible optinstall epoch homepage backup pkgrel mask pac_functions 2> /dev/null
+    unset name repology pkgver epoch url depends makedepends breaks replace gives pkgdesc hash optdepends ppa arch maintainer pacdeps patch PACPATCH NOBUILDDEP provides incompatible optinstall epoch homepage backup pkgrel mask pac_functions repo 2> /dev/null
     unset -f pkgver post_install post_remove pre_install prepare build package 2> /dev/null
     sudo rm -f "${pacfile}"
 }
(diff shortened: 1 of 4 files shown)


misc/scripts/install-local.sh
```diff
@@ -800,6 +800,9 @@ if [[ -n $pacdeps ]]; then
         touch "/tmp/pacstall-pacdeps-$i"
 
         [[ $KEEP ]] && cmd="-KPI" || cmd="-PI"
+        if pacstall -S "${i}@${REPO}"; then
```
Elsie19, line 803: This should be redirecting output.

Elsie19 (changes requested):

misc/scripts/install-local.sh
D-Brox: Oh, I forgot

misc/scripts/install-local.sh
```diff
@@ -800,6 +800,9 @@ if [[ -n $pacdeps ]]; then
         touch "/tmp/pacstall-pacdeps-$i"
 
         [[ $KEEP ]] && cmd="-KPI" || cmd="-PI"
+        if pacstall -S "${i}@${REPO}" 2&>/dev/null; then
```
Elsie19, line 803:
```suggestion
        if pacstall -S "${i}@${REPO}" &>/dev/null; then
```

Elsie19 (changes requested):

Elsie19 (approved):

PR #996, opened 2023-10-23
feat!(install-local): run build inside an isolated env

D-Brox:
## Purpose

`pacscripts` can be arbitrary code, and therefore can be unsafe.

This PR reduces the area of attack by sourcing the pacscripts  and running it's functions in an isolated environment by using bwrap. This is done by only giving read/write permissions inside the `$SRCDIR` source directory and the `$pkgdir` staging directory, while giving read-only to `/`, and no  internet access. After bwrap finishes executing, it will also kill with SIGKILL any child processes that are invoked and outlast it's parent.

## Approach

Created a new file called `bwrap.sh`, providing `safe_source()`, which sources a trimmed down env file created through bwrap, `bwrap_pkgver()` and `bwrap_function()`, which run the functions inside bwrap.

## Addendum

This is a breaking change, as `sudo` must be removed from every pacscript.

Closes #691

## Checklist

- [x] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

PR #992, opened 2023-10-12
fix(install-local.sh): chown pacscript

Zahrun:
## Purpose

I often use virtual machines to test out pacscripts, for that I use a Shared Folder between the host and the guest.
When I get into that directory in the guest, the files belong to root:vboxsf.
When I do a local pacstall install, I get a permission error.
image

## Approach

After copying the pacscript to /tmp, make sure it belongs to the pacstall user

## Checklist

- [X] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

PR #1004, opened 2023-11-12
fix(install-local): ensure dirs made by sudo are readable

srd424:
## Purpose

I run my system with a 0007 umask, which means some of the files and directories created by pacstall with sudo aren't then readable by it when it returns back to running as a normal user.

## Approach

Simply chmod the files / directories after they are created. There doesn't seem a way to change the umask of sudo-invoked processes easily, looks like that would require modifying sudoers - that might be worth considering in future if it turns out there are many other places that need fixing up.

## Progress

This patch fixes the two locations I've found so far, but I suspect there are others. It's good enough to allow me to install the package I wanted :)

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -716,6 +716,7 @@ homedir="$(eval echo ~"$PACSTALL_USER")"
 export homedir
 
 sudo cp "${PACKAGE}.pacscript" /tmp
+sudo chmod a+r "/tmp/${PACKAGE}.pacscript"
 pacfile="$(readlink -f "/tmp/${PACKAGE}.pacscript")"
 export pacfile
 mapfile -t FARCH < <(dpkg --print-foreign-architectures)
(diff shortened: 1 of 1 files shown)


oklopfer:
@Elsie19 this is a better and more thorough solution than #992 was; this one should probably get merged.

D-Brox:
This may still be needed tho, @oklopfer

@srd424 Can you check out #996? We are moving to isolated builds and I'd like to see if this needs to be added to it too.

Be aware you must remove sudo calls from the pacscripts before installing from that patch.

Kjir:
FYI: This issue is blocking in my case because my root user (for reasons unbeknownst to me, the root user has `umask 0077` on my work machine).
I think it would be wise to merge this quick fix, and the code can be further adjusted in #996 when it it clear how to proceed.

Elsie19:
> FYI: This issue is blocking in my case because my root user (for reasons unbeknownst to me, the root user has `umask 0077` on my work machine).
> I think it would be wise to merge this quick fix, and the code can be further adjusted in #996 when it it clear how to proceed.

Discussing with the team right now.

srd424:
Ah, apologies for not having found the time to look at #996 yet, things have been crazy here. It definitely sounds like a good idea!

oklopfer (approved):

PR #1012, opened 2023-12-10
feat: add `priority` key

Elsie19:
## Purpose

It allows more control for package maintainers.

## Addendum

~~@oklopfer wants this because it would prevent many cases of unicorn deleting itself.~~
It can be used for who knows what but it's gonna be added so yeah.

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -483,7 +483,13 @@ function makedeb() {
         deblog "Architecture" "all"
     fi
     deblog "Section" "Pacstall"
-    deblog "Priority" "${priority:-optional}"
+
+    if [[ ${priority} == "essential" ]]; then
+        deblog "Priority" "required"
+        deblog "Essential" "yes"
+    else
+        deblog "Priority" "${priority:-optional}"
+    fi
 
     if [[ $name == *-git ]]; then
         deblog "Vcs-Git" "${url}"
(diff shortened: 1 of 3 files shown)


misc/scripts/install-local.sh
```diff
@@ -785,6 +785,15 @@ if ! checks; then
     return 1
 fi
 
+# If priority exists and is required, and also that this package has not been installed before (first time)
+if [[ -n ${priority} && ${priority} == 'required' ]] && ! is_package_installed "${name}"; then
```
D-Brox, line 789: Wouldn't this be enough?
```suggestion
if [[ ${priority} == 'required' ]] && ! is_package_installed "${name}"; then
```

Elsie19: No. We need to check that `priority` exists first, not assume it does.

D-Brox: If it doesn't exist, then it is interpreted as blank, which is different from `"required"`

Elsie19: I'd like it to still be more explicit.

oklopfer:
how does this actually solve our rhino dependency problems tho, if the `-core` packages are meant to replace each other? would it not fail out when trying to swap them?

D-Brox:
@oklopfer IIRC the replaces field takes care of that. Needs testing tho

Elsie19:
Note: the key we need is `Essential` for some reason. We can have both though, it's no big deal.

misc/scripts/install-local.sh
```diff
@@ -785,6 +785,15 @@ if ! checks; then
...
+    ask "This package has 'priority=required', meaning once this is installed, it should be assumed to be uninstallable. Do you want to continue?" Y
+    if ((answer != 0)); then
+        cleanup
+        exit 1
```
oklopfer, line 793: this prompt doesnt work btw it just kills the entire script regardless of response

oklopfer (changes requested):

misc/scripts/install-local.sh
Elsie19: Huh? How?

oklopfer:
https://github.com/Elsie19/pacstall/pull/5

Elsie19:
Continued from https://github.com/Elsie19/pacstall/pull/5

oklopfer (approved):
tested LGTM 🚢

PR #1020, opened 2023-12-23
fix(install-local.sh): use `curl -sO` instead of `wget -q` for downloading patches

vigress8:
## Purpose

When downloading patches, `wget -q` includes URL parameters in the resulting filename. This causes e.g. patches from the AUR to have `?h=${pkgname}` appended to their filenames, which is ugly.

## Approach

The `-O` flag for `curl` tells it to use the remote file name, so it ignores URL parameters, solving the problem. Otherwise the functionality is identical.

## Addendum

One potential issue is that the `--output-dir` flag was introduced in `curl` 7.73.0, and the Debian 10 non-backports repo (and by extension, Ubuntu 20.04) uses an older version. An easy solution is to `cd` into PACSTALL_patchesdir, download the patches and `cd` back. But I'd like to discuss which approach is best.

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -970,7 +970,10 @@ if [[ -n $patch ]]; then
     for i in "${patch[@]}"; do
         curl -sO "$i"
     done
-    popd > /dev/null || return 1
+    popd > /dev/null || {
+        fancy_message error "Could not enter into patches directory"
+        return 1
+    }
     export PACPATCH="$PWD/PACSTALL_patchesdir"
 fi
 
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -961,13 +961,13 @@ mkdir -p "${SRCDIR}"
 if [[ -n $patch ]]; then
     fancy_message info "Downloading patches"
     mkdir -p PACSTALL_patchesdir
-	# NOTE: not using --output-dir,
-	# since Buster/Focal include a version of curl w/o that option.
-	pushd PACSTALL_patchesdir > /dev/null || return 1
+    # NOTE: not using --output-dir,
+    # since Buster/Focal include a version of curl w/o that option.
+    pushd PACSTALL_patchesdir > /dev/null || return 1
     for i in "${patch[@]}"; do
         curl -sO "$i"
     done
-	popd > /dev/null || return 1
+    popd > /dev/null || return 1
     export PACPATCH="$PWD/PACSTALL_patchesdir"
 fi
 
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -963,7 +963,10 @@ if [[ -n $patch ]]; then
     mkdir -p PACSTALL_patchesdir
     # NOTE: not using --output-dir,
     # since Buster/Focal include a version of curl w/o that option.
-    pushd PACSTALL_patchesdir > /dev/null || return 1
+    pushd PACSTALL_patchesdir > /dev/null || {
+        fancy_message error "Could not enter into patches directory"
+        return 1
+    }
     for i in "${patch[@]}"; do
         curl -sO "$i"
     done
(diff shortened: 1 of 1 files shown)


Elsie19:
We only support the latest LTS so 20.04 is not our concern.

Elsie19:
Actually let me check how makepkg does it first. I'd like to use their behavior if possible as we try to take feature parity with them.

Elsie19:
Uhhhh wait they don't have `patch`... I think I gaslit myself into thinking they had it for ages 😆 ok whatever

vigress8:
I definitely think Pacstall's approach with a dedicated `patch` option is superior, by the way.

oklopfer:
@Elsie19 focal/20.04 is supported until 2025; we should not drop it so quickly. In addition, I personally use pacstall on 20.04 for Ubuntu Touch purposes regularly, so maintaining its usability is important. An alternative to @vigress8's suggestion of `cd`ing, we could just check lsb_release and continue to use `wget -q` for focal, since it seems like this isn't an issue of function, but just of looks. And we should not be breaking function for looks.

vigress8:
Then you would have to check the lsb_release inside pacscripts and modify patch filenames if `wget -q` was used. That makes no sense.
`cd`'ing would be the best bet, until focal catches up or reaches EOL

vigress8:
I actually think the very best approach would be to make `patch` behave like the `source` array in PKGBUILDs, where an optional filename may be provided.
But that's for when an equivalent of the `source` array proper is implemented in Pacstall. This works fine for now

Elsie19 (approved):
Run through shfmt (`shfmt -bn -ci -sr -s -i 4 -w misc/scripts/install-local.sh`)

vigress8:
> `shfmt -bn -ci -sr -s -i 4 -w misc/scripts/install-local.sh`

Doesn't seem to modify anything

Elsie19:
> > `shfmt -bn -ci -sr -s -i 4 -w misc/scripts/install-local.sh`
> 
> Doesn't seem to modify anything

image
It should, try again.

vigress8:
Yeah, I just forgot to switch my branch 💀

misc/scripts/install-local.sh
```diff
@@ -961,9 +961,13 @@ mkdir -p "${SRCDIR}"
 if [[ -n $patch ]]; then
     fancy_message info "Downloading patches"
     mkdir -p PACSTALL_patchesdir
+    # NOTE: not using --output-dir,
+    # since Buster/Focal include a version of curl w/o that option.
+    pushd PACSTALL_patchesdir > /dev/null || return 1
```
Elsie19, line 966: Let's add an error message on that `return 1` just so people know what went wrong if it does go wrong.

Elsie19 (changes requested):

misc/scripts/install-local.sh
vigress8: You can commit to my branch. I'm bad at messages

Elsie19: It's your PR, just write down what the error actually is.

misc/scripts/install-local.sh
```diff
@@ -961,9 +961,16 @@ mkdir -p "${SRCDIR}"
...
+    # since Buster/Focal include a version of curl w/o that option.
+    pushd PACSTALL_patchesdir > /dev/null || {
+        fancy_message error "Could not enter into patches directory"
+        return 1
+    }
     for i in "${patch[@]}"; do
-        wget -q "$i" -P PACSTALL_patchesdir
+        curl -sO "$i"
     done
+    popd > /dev/null || return 1
```
Elsie19, line 973: Add the same error for the `popd` part as well.

Elsie19 (changes requested):

Elsie19:
https://github.com/pacstall/pacstall-programs/blob/master/packages/aseprite/aseprite.pacscript

This is the only package we'll need to update.

PR #1023, opened 2023-12-28
fix(download): download literal name

Elsie19:
## Purpose

If a URL has `%20` (encoded space) in it, and the download runs with `wget`, it will download but then pacstall assumes that the downloaded file still has the encoded space, messing everything up.

## Addendum

Thanks @Mythbusters123 for bringing this to my attention!

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -321,11 +321,11 @@ function download() {
             curl -L -# -o "${1##*/}" "$1" || return 1
             ;;
         quiet-wget)
-            wget -q -- "$1" 2>&1 || return 1
+            wget -q -O "${1##*/}" -- "$1" 2>&1 || return 1
             ;;
         payload) ;;
         wget | *)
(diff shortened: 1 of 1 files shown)


Mythbusters123:
I would also like to add if the URL has anything encoded either a comma or space or whatever, then that would also be broken.

Elsie19:
Once merged, remove https://github.com/pacstall/pacstall-programs/blob/master/packages/lunarclient-app/lunarclient-app.pacscript#L1

Elsie19:
Test for discord.

D-Brox (approved):

Elsie19:
> I would also like to add if the URL has anything encoded either a comma or space or whatever, then that would also be broken.

Can you give an example?

D-Brox:
@Elsie19, he was just saying that anything that is URL-encoded (%00 to %ff) could affect the filename, not just the encoded space.

This PR fixes any of those cases tho.

PR #1024, opened 2024-01-21
fix(install-local.sh): do more repair to `FARCH` array

oklopfer:
## Purpose
`FARCH` only worked if the given pacscript's `arch` array was singular (i.e. if more than one foreign architecture was listed, then it did not work)

## Approach
Fix the check to loop properly

## Progress

- [x] Did it
- [x] Tested

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -252,7 +252,7 @@ function get_incompatible_releases() {
 }
 
 function is_compatible_arch() {
-    local input=("${@}") ret=1
+    local input=("${@}") ret=1 pacarch farch
     if [[ " ${input[*]} " =~ " any " ]]; then
         ret=0
     elif [[ " ${input[*]} " =~ " ${CARCH} " ]]; then
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -252,23 +252,25 @@ function get_incompatible_releases() {
 }
 
 function is_compatible_arch() {
-    local input=("${@}")
+    local input=("${@}") ret=1
     if [[ " ${input[*]} " =~ " any " ]]; then
-        return 0
-    elif ! [[ " ${input[*]} " =~ " ${CARCH} " ]]; then
-        if [[ -n ${FARCH[*]} ]]; then
-            if [[ " ${FARCH[*]} " =~ " ${input[*]} " ]]; then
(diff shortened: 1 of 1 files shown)


misc/scripts/install-local.sh
```diff
@@ -252,23 +252,25 @@ function get_incompatible_releases() {
...
+        for pacarch in "${input[@]}"; do
+            for farch in "${FARCH[@]}"; do
+                if [[ "${pacarch}" == "${farch}" ]]; then
+                    fancy_message warn "This package is for ${BBlue}${farch}${NC}, which is a foreign architecture"
+                    ret=0
+                fi
+            done
+        done
+    fi
+    if ((ret==1)); then
```
Elsie19, line 270:
```suggestion
    if ! ((ret)); then
```

Elsie19 (changes requested):

misc/scripts/install-local.sh
oklopfer: Doesn't work.

oklopfer: with mine:
```bash
[~] rhino-live@rhino-live $ pacstall -I aws-cli-v2-bin
(aws-cli-v2-bin) Do you want to view/edit the pacscript? [y/N] n
[+] INFO: Sourcing pacscript
[!] ERROR: This Pacscript does not work on arm64
[~] rhino-live@rhino-live $ 
```
with yours:
```bash
[~] rhino-live@rhino-live $ pacstall -I aws-cli-v2-bin
(aws-cli-v2-bin) Do you want to view/edit the pacscript? [y/N] n
[+] INFO: Sourcing pacscript
[~] rhino-live@rhino-live $ 
```

misc/scripts/install-local.sh
```diff
@@ -252,23 +252,29 @@ function get_incompatible_releases() {
 }
 
 function is_compatible_arch() {
-    local input=("${@}")
+    local input=("${@}") ret=1
```
Elsie19, line 255:
```suggestion
    local input=("${@}") ret=1 pacarch farch
```

Elsie19 (changes requested):

misc/scripts/install-local.sh
oklopfer: Why would I localize those? They’re fill-ins for a for loop.

PR #1025, opened 2024-01-21
feat(install-local.sh): add `DCPUFLAGS` + `cpuflags` arrays

oklopfer:
## Purpose
adds internal cpuflag check for making a pacscript rely on it or choose a package based on certain flags

## Approach

add 2 new arrays

## Progress

- [x] Do it
- [x] Test it

## Checklist

### DIFF
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -281,7 +281,7 @@ function is_compatible_cpuflags() {
                 break
             fi
         done
-        if ((flag_found==0)); then
+        if ((flag_found == 0)); then
             ret=1
             fancy_message error "This Pacscript requires the CPU flag: ${BBlue}${pcpuflag}${NC}, but your CPU does not support it."
             break
--- a/misc/scripts/install-local.sh
+++ b/misc/scripts/install-local.sh
@@ -272,7 +272,7 @@ function is_compatible_arch() {
 }
 
 function is_compatible_cpuflags() {
-    local input=("${@}") ret=0
+    local input=("${@}") ret=0 pcpuflag dcpuflag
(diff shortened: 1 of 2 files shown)


misc/scripts/install-local.sh
```diff
@@ -271,6 +271,25 @@ function is_compatible_arch() {
...
+    local input=("${@}") ret=0
+    for pcpuflag in "${input[@]}"; do
+        local flag_found=0
+        for dcpuflag in "${DCPUFLAGS[@]}"; do
+            if [[ "$pcpuflag" == "$dcpuflag" ]]; then
+                flag_found=1
+                break
+            fi
+        done
+        if ((flag_found==0)); then
```
Elsie19, line 284:
```suggestion
        if ((flag_found)); then
```

Elsie19 (changes requested):

misc/scripts/install-local.sh
oklopfer: Nope. Not correct.

Elsie19: Why not.

oklopfer: This would be checking for the existence of the variable, not the value of it, 4head.

Elsie19: no it wouldn't... `0` evaluates to success and `1` to fail, then put that in an if statement and it's checking exit codes.

oklopfer: Except it doesn’t. Try it yourself.

Elsie19: image

Elsie19: Fine I concede bash is being stupid.

PR #1030, opened 2024-01-27
chan(install.sh): remove `-qq` from most `apt` invocations

vigress8:
Downloading a potentially large amount of data quietly makes the install look like it hangs. There's no good reason to hide this information from the user. The warning about the CLI interface may be annoying, but it's much less of a UX issue than downloading hundreds of megabytes without any clear indication of doing so.

The very first invocation is more debatable to me, since it's more of a check than anything else. However, I also question the need to install both `curl` and `wget` during the installation - one of them is entirely sufficient.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -101,7 +101,10 @@ fi
 echo
 if [[ -z "$(find -H /var/lib/apt/lists -maxdepth 0 -mtime -7)" ]]; then
     fancy_message info "Updating"
-    apt-get update
+    case "${GITHUB_ACTIONS}" in
+        true) apt-get update -qq ;;
+        *) apt-get update ;;
+    esac
 fi
 
 fancy_message info "Installing packages"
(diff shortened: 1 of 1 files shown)


D-Brox (dismissed):

Elsie19:
We have it quiet so that the CI logs don't balloon in size.

D-Brox:
> We have it quiet so that the CI logs don't balloon in size.

@Elsie19 then only run quiet in the CI, duh

Elsie19:
> > We have it quiet so that the CI logs don't balloon in size.
> 
> @Elsie19 then only run quiet in the CI, duh

Elsie19 (changes requested):
Make it quiet if `GITHUB_ACTIONS` variable is set.

oklopfer (approved):
works merging

oklopfer:
just kidding I always triple check with @Elsie19

Elsie19 (approved):

PR #1038, opened 2024-02-08
feat(*-local.sh)!: `${url}` -> `${source[@]}`

oklopfer:
## Purpose

PKGBUILD parity!!

### Change Log:
- `url` variable converted to `source` array
    - `-deb` packages/`*.deb` entries can only be provided as a singular source
- `hash` variable converted to `hash` array
    - Entries can be skipped with `SKIP`
- Added support for `source_${CARCH}` + `hash_${CARCH}`
    - Deprecates external `CARCH` condition checks in favor of using `hash_arm64=()` + `hash_amd64=()`, for example
    - Entries are appended to regular `source` and `hash` arrays
        - An internal check is done to only append the native architecture
        - This means a package could entirely omit using `source` as long as it includes at least one `source_${CARCH}` option.
    - While `.deb` files can only be provided as a singular source, if two different `.deb` files are provided architecture independently, then the script will still work correctly
- Download made quiet by default, but `fancy_message` info provided
    - Use `PACSTALL_DOWNLOADER=verbose-{method}` for verbose output
- Clone info made quiet, but `fancy_message` info provided
- Added support for more extraction methods (like `rar` and `7z`)
-  Improved `tar` extraction methods for supported `tar.*` extensions
- Added support for new `source` entry syntax options:
    - `file://` syntax
        - For installing local archives relative to the pacscript
        - Generally will look like `file://${PWD}/example.zip` or  `file:///home/pacstall/example.zip`
    - `git+` syntax
        - For git sources that do not have `*.git`
    - `#branch=`, `#tag=`, and `#commit=` syntax
        - For specifying specific commits to clone from for git sources
    - `custom-location::` syntax
        - For setting custom names for source entry downloads
            - Does not change an archive's name after extraction, only before
    - `?*` auto-removal syntax
        - For dropping URL headers from destination filenames 
    - A full link is ordered: `custom-location.zip::git+file://example.com/archive.tar.xz?h=file#branch=master`
        - Extraction method is calculated from `url` and not `dest` so that an example like above would still extract correctly
- New exposed variables: 
    - `pkgbase`
        - Always dependent on the first `source` entry
        - All subsequent source entries will be placed inside this directory
        - If the first entry is an archive to be extracted or a cloned git repository, then this will be the entered directory
            - If using `PACSTALL_PAYLOAD`, this will be `/tmp/pacstall-pacdep/${entered_directory}`, as is already the case
            - Otherwise, this will be `/tmp/pacstall/${entered_directory}`, as is already the case
        - If no directory is entered, then this will become `/tmp/pacstall/$PACKAGE~$pkgver` and all source items, including the first one, will be placed here
            - This also applies for non-extracted archives
            - If using `PACSTALL_PAYLOAD`, this is instead still `/tmp/pacstall-pacdep`
    - `git_pkgver`
        - Only exposed if the package name is `-git`
        - Assumes the first `source` entry will always be a git repository
        - Provides the short-form hash of the first source's git commit, as used in the final `pkgver`
- New usable array for pacscripts: 
    - `noextract`
        - Used like: `noextract=("custom-location.zip")`
- New internal functions helpful for pacscripts: 
    - `parse_source_entry`
        - Outputs variables `dest` + `url`, as well as `git_branch`/`git_tag`/`git_commit` if they apply (but they are mutually exclusive)
        - Used like: `parse_source_entry "${source[4]}"` for the 5th source item, for example
     - `calc_git_pkgver` 
         - Outputs variable `comp_git_pkgver`
         - Depends on `parse_source_entry`
         - Used like: 
         `parse_source_entry "${source[4]}" && comp_git_pkgver` for the 5th source item, for example
- Deprecated `pkgver()` in favor of internalizing with `calc_git_pkgver`
- Git integrity checks improved by confirming cloned commit `hash` is same as `comp_git_pkgver`
- Maintaining functionality of current pacscripts ensured, conversion could be as simple as:
    - Changing `url` variable to `source` array
    - Changing `hash` variable to `hash` array
    - For `-git` packages:
        - Remove `pkgver()` usage
        - For any packages not containing `*.git` in their url, add `git+` in front of `https://`
- The majority of the `function` declarations in `install-local.sh` have been moved
    - Sourcing `install-local.sh` now chain sources `download-local.sh`, which sources `build-local.sh`
    - This creates 3 files of around 500 lines, which should make parsing them much easier
        - This is doubly improved, as GitHub now also properly finds and registers their symbols
    - The `log` function was renamed to `meta_log`
        - This is the only function that was renamed
        - It was not used anywhere else except `install-local.sh`
        - Calling it `log` was an oversimplification 
        - It is now significantly easier to find the mentions of the function
        - It is now accurately named to what it does: log package metadata

here's a test pacscript:
```bash
name="renpy-full"
gives="renpy"
replace="${gives}"
breaks=("${gives}7" "${gives}7-full" "${gives}-nightly")
repology=("project: ${gives}")
pkgdesc="A visual novel engine"
pkgver="8.1.3"
arch=("amd64" "arm64")
homepage='https://www.renpy.org'
maintainer="vigress8 <vig@disroot.org>"

noextract=("${gives}-${pkgver}-sdk.tar.bz2")

source=(
  "custom-${gives}-steam.zip::https://www.renpy.org/dl/${pkgver}/${gives}-${pkgver}-steam.zip"
  "custom-${gives}-web.zip::https://www.renpy.org/dl/${pkgver}/${gives}-${pkgver}-web.zip"
  "${gives}-pygame::git+https://github.com/renpy/pygame_sdl2.git"
  "pygame-web::https://github.com/renpy/pygame_sdl2.git"
  "git+https://github.com/rhino-linux/rhino-pkg"
  "oklopfer::https://github.com/oklopfer/rhino-pkg.git"
  "https://www.renpy.org/dl/${pkgver}/${gives}-${pkgver}-sdk.tar.bz2"
)

hash=(
  "d7470f561fcdc277087a0d86b01d6f73caced8a8d302e66dc494060f536d145c"
  "351e7a9680f3b972f6bff67bd5952558091d80ffd05fbca295ff90aa25a4f144"
  "SKIP"
  "SKIP"
  "SKIP"
  "SKIP"
  "b15fdaf1322f18aaa1a588761f52543683aefd1a3c9e45c5bf045e3a4ad468b7"
)

package() {
  echo "${makedepends[*]}"
  echo "${PWD}" & ls
}
```

Conversion script for `pacstall-programs`:
```python
#!/usr/bin/env python3

import re

def subst(file, search_p, replace_p):
    with open(file, 'r') as f:
        contents = f.readlines()

    with open(file, 'w') as f:
        for line in contents:
            line = re.sub(search_p, replace_p, line)
            f.write(line)


def delete_block(file, begin, end):
    with open(file, 'r') as f:
        contents = f.readlines()

    begin_index = None
    end_index = None
    for index, line in enumerate(contents):
        if re.search(begin, line):
            begin_index = index
        if re.search(end, line):
            end_index = index
            break

    if begin_index is not None and end_index is not None:
        del contents[begin_index : end_index + 1]
    with open(file, 'w') as f:
        f.writelines(contents)


with open("packagelist", 'r') as plist:
    for pname in plist:
        p = pname.rstrip()
        ppath = f"packages/{p}/{p}.pacscript"
        if re.search('-git', p):
            delete_block(ppath, r'^pkgver\(\)', '^}')
            subst(ppath, r'url="https(.*)(?<!\.git)"', r'url="git+https\1"')

        subst(ppath, r'url=(.*)', r'source=(\1)')
        subst(ppath, r'hash=(.*)', r'hash=(\1)')
```

## Progress

- [x] Switch to arrays
- [x] Make them work
- [x] Give them awesome syntax options

## Checklist


misc/scripts/install-local.sh
```diff
@@ -99,7 +99,7 @@ function log() {
             echo "_homepage=\"${homepage}"\"
         fi
         if [[ -z $gives && $name == *-deb ]]; then
-            echo "_gives=\"$(dpkg -f ./"${url##*/}" | sed -n "s/^Package: //p")"\"
+            echo "_gives=\"$(dpkg -f ./"${source##*/[0]}" | sed -n "s/^Package: //p")"\"
```
oklopfer, line 102: This ain’t right

Elsie19 (changes requested):

misc/scripts/install-local.sh
```diff
@@ -905,52 +905,235 @@ if ! is_package_installed "${name}"; then
...
+                return 1
+            fi
         fi
     fi
-fi
+}
+
+function genextr_declare() {
+    ext_method=""
+    ext_dep=""
```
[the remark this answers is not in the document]
oklopfer: no it will die

misc/scripts/checks.sh
```diff
@@ -136,11 +136,58 @@ function lint_version() {
...
+        if [[ ${file_name} == *.deb ]]; then
+            fancy_message error ".deb files can only be provided as a singular 'source'"
+            ret=1
+            break
+        fi
+    done
+}
+
+function lint_source() {
+    local ret=0 test_source has_source=0 known_archs=("amd64" "arm64" "armel" "armhf" "i386" "mips64el" "ppc64el" "riscv64" "s390x")
```
Elsie19, line 154: `known_archs` can be replaced with:

```bash
local known_archs=()
mapfile -t known_archs < <(dpkg-architecture --list-known)
```

misc/scripts/checks.sh
```diff
@@ -282,12 +329,28 @@ function lint_replace() {
 }
 
 function lint_hash() {
-    local ret=0
+    local ret=0 test_hash known_archs=("amd64" "arm64" "armel" "armhf" "i386" "mips64el" "ppc64el" "riscv64" "s390x")
```
Elsie19, line 332: `known_archs` can be replaced with:

```bash
local known_archs=()
mapfile -t known_archs < <(dpkg-architecture --list-known)
```

Elsie19 (changes requested):

misc/scripts/install-local.sh
```diff
@@ -125,6 +123,56 @@ function log() {
...
+        dest="${entry%%::*}"
+    elif [[ $entry != *#*=* ]]; then
+        url="$entry"
+        dest="${url##*/}"
+    fi
+}
+
+function calc_git_pkgver() {
+    unset comp_git_pkgver
+    local calc_commit
```
Elsie19, line 160:
```suggestion
```

misc/scripts/install-local.sh
```diff
@@ -125,6 +123,56 @@ function log() {
...
+    if [[ -n ${git_branch} ]]; then
+        calc_commit="git ls-remote ${url} ${git_branch}"
+    elif [[ -n ${git_tag} ]]; then
+        calc_commit="git ls-remote ${url} ${git_tag}"
+    elif [[ -n ${git_commit} ]]; then
+        calc_commit="echo ${git_commit}"
+    else
+        calc_commit="git ls-remote ${url} HEAD"
+    fi
+    comp_git_pkgver="$(${calc_commit} | cut -f1 | cut -c1-8)"
```
Elsie19, line 173:
```suggestion
    if [[ -n ${git_branch} ]]; then
        : "git ls-remote ${url} ${git_branch}"
    elif [[ -n ${git_tag} ]]; then
        : "git ls-remote ${url} ${git_tag}"
    elif [[ -n ${git_commit} ]]; then
        : "echo ${git_commit}"
    else
        : "git ls-remote ${url} HEAD"
    fi
    comp_git_pkgver="$(${_} | cut -f1 | cut -c1-8)"
```

If you don't want to merge this that's fine, but this *would* lower character count.

Elsie19 (changes requested):

misc/scripts/install-local.sh
```diff
@@ -498,7 +548,7 @@ function makedeb() {
     fi
 
     if [[ $name == *-git ]]; then
-        deblog "Vcs-Git" "${url}"
+        deblog "Vcs-Git" "${source[0]}"
```
Elsie19, line 551: Is this known/mandated?

misc/scripts/install-local.sh
```diff
@@ -745,8 +795,8 @@ fi
 
 # Running `-B` on a deb package doesn't make sense, so let's download instead
 if ((PACSTALL_INSTALL == 0)) && [[ ${name} == *-deb ]]; then
-    if ! download "${url}"; then
-        fancy_message error "Failed to download '${url}'"
+    if ! download "${source[0]}"; then
+        fancy_message error "Failed to download '${source[0]}'"
```
Elsie19, line 799: We need to check all the sources, not just the first.

Elsie19 (changes requested):

misc/scripts/install-local.sh
oklopfer: debs will always only be singular source, no issue here 👍

misc/scripts/install-local.sh
oklopfer: it will be

misc/scripts/install-local.sh
```diff
@@ -964,144 +1278,38 @@ fi
...
-        fancy_message info "Cleaning up"
-        cleanup
-        exit 1
-    }
-    # Check the integrity
-    git fsck --full || return 1
-else
+for i in "${!source[@]}"; do
+    parse_source_entry "${source[$i]}"
+    expectedHash="${hash[$i]}"
```
Elsie19, line 1283: These should be unset after.

Elsie19 (changes requested):

misc/scripts/checks.sh
```diff
@@ -136,11 +126,70 @@ function lint_version() {
     return "${ret}"
 }
 
-function lint_url() {
-    local ret=0
-    if [[ -z $url ]]; then
-        fancy_message error "Package does not contain 'url'"
+function lint_source_deb_test() {
+    # shellcheck disable=SC2206
+    local input_source=($1)
```
Elsie19, line 131:
```suggestion
    local input_source=("$@")
```

misc/scripts/checks.sh
```diff
@@ -136,11 +126,70 @@ function lint_version() {
...
+            if [[ -n ${!source_arch} ]]; then
+                test_source=()
+                if [[ -n ${source[0]} ]]; then
+                  # shellcheck disable=SC2206
+                  test_source+=(${source[*]})
+                fi
+                # shellcheck disable=SC2206
+                test_source+=(${!source_arch})
+                if [[ -n ${test_source[1]} ]]; then
+                    lint_source_deb_test "${test_source[*]}"
```
Elsie19, line 183:
```suggestion
                    lint_source_deb_test "${test_source[@]}"
```

misc/scripts/checks.sh
```diff
@@ -136,11 +126,70 @@ function lint_version() {
...
+                    if ((ret == 1)); then
+                        break
+                    fi
+                fi
+            fi
+        done
+        if [[ -n ${source[1]} ]]; then
+           lint_source_deb_test "${source[*]}"
```
Elsie19, line 191:
```suggestion
           lint_source_deb_test "${source[@]}"
```

Elsie19 (changes requested):

misc/scripts/install-local.sh
```diff
@@ -905,52 +972,335 @@ if ! is_package_installed "${name}"; then
...
+    fi
+}
+
+function file_down() {
+    fancy_message info "Copying local archive ${BPurple}${dest}${NC}"
+    cp -r "${url}" "${dest}" || fail_down
+    genextr_declare
+    if [[ -n ${ext_method} ]]; then
+        genextr_down
+    fi
```
oklopfer, line 1252: I need to correct this so that elif [[ source[i] == source[0] && -d $dest ]]; then cd $dest && export pkgbase=$PWD; fi 

and then maybe end it with a gather_down, I have to think if that's necessary

Elsie19 (changes requested):
https://github.com/pacstall/pacstall/pull/1038/checks?check_run_id=21635631179

Add `# shellcheck ignore=CODE` on all these, but if you have any doubts about one just ping me.

misc/scripts/download-local.sh
```diff
@@ -0,0 +1,543 @@
...
+        *.lz4)
+            ext_method="lz4 -d"
+            ext_deps=("liblz4-tool")
+            ;;
+        *.tar)
+            ext_method="tar -xf"
+            ext_deps=("tar")
+            ;;
+    esac
+}
```
Elsie19, line 149: Ok so I understand why you're doing this but I don't like conditionally adding dependencies to a deb based on if someone doesn't have that compression tool installed. I know that arch uses `bsdtar` for all their archiving purposes, so we could use that by just adding `libarchive-tools` to the install process to install it.

Elsie19 (changes requested):

misc/scripts/download-local.sh
vigress8: > but I don't like conditionally adding dependencies to a deb based on if someone doesn't have that compression tool installed

The dependencies are added in any case if the archive format requires them. Doesn't matter if they're already installed or not

oklopfer: Correct, what vig said

PR #1043, opened 2024-02-23
fix({download,install}-local.sh): use more `clean_fail_down`

vigress8:


oklopfer (approved):
LGTM, @Elsie19 approbe too?

PR #1045, opened 2024-02-24
chan(*)!: `SRCDIR` -> `PACDIR`, `pkgbase` -> `srcdir`

oklopfer:
## Purpose

`pkgbase` usage is actually for multiple packages with the same source script, not for singular packages with multiple source entries in the same script. (see: https://archlinux.org/pacman/PKGBUILD.5.html#_package_splitting) This needs `package_foo()`/`build_foo()`/`prepare_foo()` syntax and a few other things, so it will be further down the line.

what I created as `pkgbase` should actually have been `srcdir`, but this means replacing `SRCDIR` with a new name.

## Approach

`SRCDIR` -> `PACDIR`
`srcdir` -> `pacdir`
`pkgbase` -> `srcdir`
`pkgbase` to be reintroduced at a later point.

## Progress

- [x] Do it
- [x] Test it

## Checklist

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -117,7 +117,7 @@ apt-get install -qq -y curl wget build-essential unzip git zstd iputils-ping lsb
 
 LOGDIR="/var/lib/pacstall/metadata"
 STGDIR="/usr/share/pacstall"
-SRCDIR="/tmp/pacstall"
+PACDIR="/tmp/pacstall"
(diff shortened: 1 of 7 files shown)


vigress8:
LGTM 🚀🏳️‍⚧️

oklopfer:
works as far as I can tell

PR #1062, opened 2024-03-29
fix(*-local.sh)!: mark `pacdeps` as deps for `-deb` pacs

oklopfer:
## Purpose

if a `-deb` script defined a pacdep, or a dependency that the deb did not think it needs, it wasn't actually marked as a dependency by dpkg. 

Also, downloading a deb with -B needed verbosity.

## Approach

Unpack the deb in a temp location, parse the control file for the Depends line, create an array, insert pacdeps, depends, and makedepends into the array, print it back out into the control file, repack the deb, install it.

also give downloading them with -B verbosity

## Progress

- [x] do it
- [x] test it
- [x] test it more

## Checklist

### DIFF
--- a/misc/scripts/build-local.sh
+++ b/misc/scripts/build-local.sh
@@ -510,14 +510,12 @@ function repacstall() {
         depends_array=("${depends_array[@]/# /}")
         depends_array=("${depends_array[@]/% /}")
     fi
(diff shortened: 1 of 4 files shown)


oklopfer:
works

PR #1064, opened 2024-03-30
fix(download-local.sh): be verbose about pacstall hooks for -debs

oklopfer:
## Purpose

pacstall hooks are different from debian hooks when in the case of `-deb`s. 

## Approach

distinguish them with verbosity. Also fix missing post_upgrade hook. Also make final cleanup terminology equal for `-debs` to how it was for everyone else (using `fancy_message sub` and announcing package finished installing)

## Progress

- [x] do it
- [x] test it as always

## Checklist


oklopfer:
Screenshot 2024-03-30 at 3 07 23 AM
gorgeous.

Elsie19 (approved):

PR #1068, opened 2024-04-01
feat!(*): run build inside an isolated env

### LINKED ISSUE #691: consider new approach to managing privileges, preferring root invocation
# Problem

I have noticed a particular design pattern common for AUR helpers, which appears also to have been adopted by the Pacstall utility.

The utility prefers to run as the regular user, and during its operation acquire permissions for elevated privileges, to be utilized when the built packages are ready for installation. 

The approach carries several limitations, including the following:

- The build work is performed and saved under the user directory, even though it has system-wide scope. Users may not want the work saved in their own space, and may benefit from invoking operations that share artifacts with operations invoked by other users.
- Users are accustomed to invoking package management tools with elevated privileges, and may include such invocations within operations already carrying  such privileges. 

# Proposal

An alternative approach would be the creation of a special user for Pacstall work operations, which carries file permissions to a designated system directory. The utility then may delegate work operations to a worker process with the reduced priviliges of this special user, and then, when the process completes, proceed with package installation.

Inverting the point at which privileges are changed, and in which direction, may  represent overall a design that is more elegant and better integrated.

# Notes

I have recently received my first exposure to Arch-family operating systems, and only a short time after learning about AUR, and wondering why the concept has not been adopted in other Linux families, came to learn of the Pacstall project. I am happy about this work, since I have been a long-time user of Ubuntu, and more recently Linux Mint, and have negotiated all the obstacles related to package management familiar to so many.

oklopfer:
fork of https://github.com/pacstall/pacstall/pull/996

## Purpose

`pacscripts` can be arbitrary code, and therefore can be unsafe.

This PR reduces the area of attack by sourcing the pacscripts  and running it's functions in an isolated environment by using bwrap. This is done by only giving read/write permissions inside the `$SRCDIR` source directory and the `$pkgdir` staging directory, while giving read-only to `/`, and no  internet access. After bwrap finishes executing, it will also kill with SIGKILL any child processes that are invoked and outlast it's parent.

## Approach

Created a new file called `bwrap.sh`, providing `safe_source()`, which sources a trimmed down env file created through bwrap, and `bwrap_function()`, which run the functions inside bwrap.

## Addendum

This is a breaking change, as `sudo` must be removed from every pacscript.

Closes #691

## Checklist


misc/scripts/bwrap.sh
```diff
@@ -0,0 +1,111 @@
...
+    mkdir /tmp/pacstall 2>/dev/null
+    safeenv="$(sudo mktemp -p "${PACDIR}")"
+    sudo chmod +r "$safeenv"
+    bwrapenv="$(sudo mktemp -p "${PACDIR}" -t "bwrapenv.XXXXXXXXXX")"
+    sudo chmod +r "$bwrapenv"
+    export bwrapenv
+    export safeenv
+
+    tmpfile="$(sudo mktemp -p "${PACDIR}")"
+    local allsource allsums src sum a_sum known_hashsums_src=("b2" "sha512" "sha384" "sha256" "sha224" "sha1" "md5") known_archs_src=("amd64" "arm64" "armel" "armhf" "i386" "mips64el" "ppc64el" "riscv64" "s390x")
```
Elsie19, line 36: `known_archs_src` should come from, you guessed it, `dpkg-architecture --list-known`.

Elsie19 (changes requested):

PR #1070, opened 2024-04-03
fix(dep-tree.sh): expand classes

oklopfer:
## Purpose

if a package is both has pacdeps and is itself a pacdep, it should be called lower than packages that have pacdeps and themselves are not already pacdeps

## Approach

expand the groups to the following priority, in descending order (1 is highest priority):

```
1)  pacdeps  +  !is_pacdep  +  !depends
    pacdeps  +  !is_pacdep  +   depends
-----------------------------------------
2) !pacdeps  +  !is_pacdep  +  !depends
-----------------------------------------
3) !pacdeps  +  !is_pacdep  +   depends
-----------------------------------------
4)  pacdeps  +   is_pacdep  +  !depends
    pacdeps  +   is_pacdep  +   depends
-----------------------------------------
5) !pacdeps  +   is_pacdep  +  !depends
-----------------------------------------
6) !pacdeps  +   is_pacdep  +   depends
```

## Progress

- [x] do it
- [x] test it

## Checklist

### DIFF
--- a/misc/scripts/dep-tree.sh
+++ b/misc/scripts/dep-tree.sh
@@ -83,41 +83,51 @@ function dep_tree.load_traits() {
 
 function dep_tree.sort_traits_into_array() {
     local pkg="${1:?No pkg given to dep_tree.sort_traits_into_array}"
-    local -n trait c_one c_two c_three c_four
+    local -n trait c_one c_two c_three c_four c_five c_six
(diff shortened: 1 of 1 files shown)


Elsie19 (changes requested):
Can you add comments to explain the different buckets? I didn't do that originally but I think we should now.

PR #1071, opened 2024-04-04
fix(upgrade.sh)!: repair old REPOS code

oklopfer:
## Purpose

when a package's `_remoterepo` is no longer existent in the `pacstallrepo` file during `-Up`, we are currently removing the repo link from the metadata of the package. This is bad practice, as in the case of simply switching to an updated repo (like `5.0.0-master`), packages are wiped from being updatable until re-installing.

## Approach

- Fix up ancient code regarding `REPOS` + `remoterepo` that was written by someone who was not paying any attention to what they were doing. 
- When a repo mismatch happens, now rather than removing their remoterepo entirely, we are replacing it with the string `_remoterepo="orphan"`, which gives the packages special treatment separate from a `local` install, where they can still be found by `-Up`.
- Also, if a repo was commented out in the `pacstallrepo` file, it would cause `-Up` to break as well - fix that by simply ignoring ones that start with `#`.
- Also, `-git` packages were erroneously getting special treatment, which caused them to have issues compared to the rest of packages for updating. Fixed that, and confirmed they work correctly now.
- Also, if repo is github or gitlab, display branch in outputs. For notification of orphaning, branch info will always be shown. For showing a valid update, branch info will be shown if branch is not `master` or `main`, otherwise it will just show same as before. 

## Progress

- [x] Do it
- [x] Test it

## Checklist

### DIFF
--- a/misc/scripts/search.sh
+++ b/misc/scripts/search.sh
@@ -209,14 +209,24 @@ elif [[ -n $UPGRADE ]]; then
     REPOS=()
     # Return list of repos with the package
     for IDX in $IDXSEARCH; do
-        ! [[ " ${REPOS[@]} " =~ " ${URLLIST[$IDX]} " ]] && mapfile -t -O"${#REPOS[@]}" REPOS <<< "${URLLIST[$IDX]}"
+        ! array.contains REPOS "${URLLIST[$IDX]}" && mapfile -t -O"${#REPOS[@]}" REPOS <<< "${URLLIST[$IDX]}"
     done
     export REPOS
     return 0
 # Check if its being used for search
 elif [[ -n $SEARCH ]]; then
     for IDX in $IDXSEARCH; do
-        echo -e "$GREEN${PACKAGELIST[$IDX]} $PURPLE@ $CYAN$(parseRepo "${URLLIST[$IDX]}") $NC"
+        searchedrepo="$(parseRepo "${URLLIST[$IDX]}")"
+        if [[ ${URLLIST[$IDX]} == *"github"* ]]; then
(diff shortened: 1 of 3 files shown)


oklopfer:
Screenshot 2024-04-04 at 3 30 42 AM

D-Brox:
This happened back in August 2021, introduced in #163 for version 1.6

We had reasons to do it that way (looking back, it was probably a bad reason tbh)

Elsie19:
> This happened back in August 2021, introduced in #163 for version 1.6
> 
> We had reasons to do it that way (looking back, it was probably a bad reason tbh)

It was for git packages iirc.

misc/scripts/search.sh
```diff
@@ -143,7 +143,7 @@ while IFS= read -r URL; do
             || fancy_message warn "Replace '~' with the full home path on \e]8;;file://$STGDIR/repo/pacstallrepo\a$CYAN$STGDIR/repo/pacstallrepo$NC\e]8;;\a"
         URL="${URL/'~'/$HOME}"
     fi
-    if ! check_url "${URL}/packagelist"; then
+    if [[ ${URL} != "#"* ]] && ! check_url "${URL}/packagelist"; then
```
Elsie19, line 146: This would fail if for some reason the string was `   # URL`.

Elsie19 (changes requested):

misc/scripts/search.sh
oklopfer: That's the entire purpose of this. Skipping commented out REPOS from the file.

Elsie19: I mean there should allow an arbitrary amount of whitespace before the `#`.

PR #1079, opened 2024-04-18
fix(install-local): reenable prompts on pacdeps when prompts are not disabled

D-Brox:
## Purpose

At the moment we disable prompts on pacdeps regardless of the `-P` flag. We should only disable when the flag is used, as this can be a security risk.

## Approach

Check the `$DISABLE_PROMPTS` variable when setting the pacdep installation command.

## Checklist


oklopfer:
tested and works, @Elsie19 all good 4 merge?

Elsie19 (approved):

PR #1080, opened 2024-04-19
fix(main): rework argument parsing and handling

D-Brox:
## Purpose

We currently don't:
- deduplicate flags
- check if more than one command flag is being used
- properly parse `-h` in argument cluster
- reorder long flags options correctly

But we should.

## Approach

Fix it all

## Progress

It's done and tested

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -506,10 +506,10 @@ fi
 argument_list=()
 
 for i in "${@}"; do
-    # Just add argument if doesn't start with a single hyphen.
+    # Just add argument if doesn't start with exactly one hyphen.
     if ! [[ ${i} =~ ^-[^-] ]]; then
-        # if argument is '-P' or '-K', add to beginning of argument list.
-        if [[ ${i} == "-B" || ${i} == "-P" || ${i} == "-K" ]]; then
+        # if argument is '-B', '-P', '-K' or '-Nc', add to beginning of argument list.
+        if [[ ${i} == "--build" || ${i} == "--disable-prompts" || ${i} == "--keep" || ${i} == "--nocheck" ]]; then
             argument_list=("${i}" "${argument_list[@]}")
         else
             argument_list+=("${i}")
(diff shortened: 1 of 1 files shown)


pacstall
```diff
@@ -530,8 +531,37 @@ for i in "${@}"; do
...
+        hashed_arguments[$mapped]=0
+    fi
+done
+unset hashed_arguments
+
+# Check if only one command flag is being used
+short_commands=("-I" "-S" "-R" "-D" "-A" "-U" "-L" "-Up" "-Qi" "-T" "-V" "-h")
+long_commands=("--install" "--search" "--remove" "--download" "--add-repo" "--update" "--list" "--upgrade" "--query-info" "--tree" "--version" "--help")
+commands=("${short_commands[@]}" "${long_commands[@]}")
+matches=($(comm -12 <(sort <(printf "%s\n" "${unique_argument_list[@]}")) <(sort <(printf "%s\n"  "${commands[@]}"))))
```
oklopfer, line 556: can you explain what im looking at here please

D-Brox: Compare the list of deduplicated args with the full list of command flags.

If there is more than one match, that means you are trying to run multiple commands at the same time.

pacstall
```diff
@@ -530,8 +531,37 @@ for i in "${@}"; do
...
-set -- "${argument_list[@]}"
+# Remove duplicates
+declare -A arg_map=(
+    ["--install"]="-I"     ["--search"]="-S"   ["--remove"]="-R"          ["--download"]="-D"
+    ["--add-repo"]="-A"    ["--update"]="-U"   ["--list"]="-L"            ["--upgrade"]="-Up"
+    ["--query-info"]="-Qi" ["--tree"]="-T"     ["--version"]="-V"         ["--help"]="-h"
+    ["--build"]="-B"       ["--keep"]="-K"     ["--disable-prompts"]="-P" ["--nocheck"]="-Nc"
+)
+declare -A hashed_arguments
+unique_argument_list=()
```
oklopfer, line 542: you can (and maybe should) `unset unique_arguments_list` here instead of setting an empty array. When you do `+=($arg)`, the first one will automatically initialize it.

PR #1101, opened 2024-05-06
fix(bwrap): dns resolution in the sandbox

D-Brox:
## Purpose

Since we tmpfs `/run` with bwrap, dns resolution is borked. So we just need to ro-bind the `/run/systemd/resolve` dir.

## Approach

Add the flag to bwrap

## Progress

Done and tested with cargo packages

## Checklist

### DIFF
--- a/misc/scripts/bwrap.sh
+++ b/misc/scripts/bwrap.sh
@@ -117,12 +117,13 @@ EOF
     if [[ ! -d ${LOGDIR} ]]; then
         sudo mkdir -p "${LOGDIR}"
     fi
-    local share_net
+    local share_net dns_resolve
     if [[ ${external_connection} == "true" ]]; then
         share_net="--share-net"
+        dns_resolve="--ro-bind /run/systemd/resolve /run/systemd/resolve"
     fi
     sudo bwrap --unshare-all ${share_net} --die-with-parent --new-session --ro-bind / / \
-        --proc /proc --dev /dev --tmpfs /tmp --tmpfs /run --dev-bind /dev/null /dev/null \
+        --proc /proc --dev /dev --tmpfs /tmp --tmpfs /run ${dns_resolve} --dev-bind /dev/null /dev/null \
         --bind "$STAGEDIR" "$STAGEDIR" --bind "$PACDIR" "$PACDIR" --setenv LOGDIR "$LOGDIR" \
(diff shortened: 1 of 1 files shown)


misc/scripts/fetch-sources.sh
```diff
@@ -385,6 +385,16 @@ function file_down() {
     gather_down
 }
 
+function crates_down() {
+    local cargo_toml=($(find "${srcdir}" -name Cargo.toml))
+    if ((${#cargo_toml[@]} != 0)); then
+        fancy_message sub "Fetching cargo dependencies"
+        for crate in "${cargo_toml[@]}"; do
+            CARGO_HOME="${srcdir}/.cargo" cargo fetch --manifest-path="$crate" --target "$(cargo -vV | grep "host" | cut -d' ' -f2)"
```
Elsie19, line 393:
```suggestion
            CARGO_HOME="${srcdir}/.cargo" cargo fetch --manifest-path="$crate" --target "$(rustc -vV | sed -n 's|host: ||p')"
```

Elsie19 (changes requested):

misc/scripts/fetch-sources.sh
```diff
@@ -385,6 +385,16 @@ function file_down() {
     gather_down
 }
 
+function crates_down() {
+    local cargo_toml=($(find "${srcdir}" -name Cargo.toml))
```
Elsie19, line 389: Use mapfile.

Elsie19 (changes requested):

oklopfer (approved):

misc/scripts/bwrap.sh
```diff
@@ -117,12 +117,14 @@ EOF
     if [[ ! -d ${LOGDIR} ]]; then
         sudo mkdir -p "${LOGDIR}"
     fi
-    local share_net
+    local share_net dns_resolve
     if [[ ${external_connection} == "true" ]]; then
         share_net="--share-net"
+        dns_resolve="--ro-bind /run/systemd/resolve /run/systemd/resolve"
```
oklopfer, line 123: this will almost assuredly cause issues in docker containers.

oklopfer (changes requested):

oklopfer (approved):

PR #1108, opened 2024-05-07
fix({build,package}.sh): log previously installed pacdeps as such

oklopfer:
## Purpose

if I install a pacdep manually, it loses connection to its parent, and is not properly logged as a pacstall depends in metadata.

## Approach

in build.sh, check if existing metadata lists this package as a pacdep; if so, log it as one. 
in package.sh, if the package is already installed, check if it has the declaration in its metadata; if not, add it

## Progress

- [x] do it
- [x] test it

## Addendum

part of https://github.com/pacstall/pacstall-programs/pull/5986

## Checklist

### DIFF
--- a/misc/scripts/build.sh
+++ b/misc/scripts/build.sh
@@ -678,9 +678,9 @@ function repacstall() {
 }
 
 function check_if_pacdep() {
-    package="${1}"
-    finddir="${2}"
-    found="$(find "${finddir}" -type f -exec awk -v pkg="${package}" '
+    local package="${1}"
+    local finddir="${2}"
+    local found="$(find "${finddir}" -type f -exec awk -v pkg="${package}" '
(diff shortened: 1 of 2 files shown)


misc/scripts/build.sh
```diff
@@ -677,6 +677,18 @@ function repacstall() {
     install_deb
 }
 
+function check_if_pacdep() {
+    package="${1}"
+    finddir="${2}"
+    found="$(find "${finddir}" -type f -exec awk -v pkg="${package}" '
```
Elsie19, line 683:
```suggestion
    local package="${1}"
    local finddir="${2}"
    local found="$(find "${finddir}" -type f -exec awk -v pkg="${package}" '
```

Elsie19 (changes requested):

D-Brox (approved):

PR #1118, opened 2024-05-22
feat(add-repo.sh): allow `-Rr`/`--remove-repo`

oklopfer:
## Purpose

we should allow this tbh

## Approach

just look below

## Progress

- [x] allow `-A`/`--add-repo` to accept metalinks
    - `github:user/repo` (defaults to branch `master`) [also works with provider `gitlab`]
    - `github:user/repo#branch` [also works with provider `gitlab`]
    - `local:/path/to/dir`
- [x] create `-R`/`--remove-repo` that uses same file, but slight reversal process
- ~~**add alias options to call repos**~~ save for 5.2.0

## Checklist

### DIFF
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -23,6 +23,7 @@
 # along with Pacstall. If not, see <https://www.gnu.org/licenses/>.
 
 parse_repo() {
+    local ADDR
     IFS=':' read -ra ADDR <<< "$1"
     PROV="${ADDR[0]}"
     USER=$(echo "${ADDR[1]}" | cut -d'/' -f1)
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -22,7 +22,7 @@
 # You should have received a copy of the GNU General Public License
 # along with Pacstall. If not, see <https://www.gnu.org/licenses/>.
 
-parse_repo() {                                  
+parse_repo() {
     IFS=':' read -ra ADDR <<< "$1"
     PROV="${ADDR[0]}"
     USER=$(echo "${ADDR[1]}" | cut -d'/' -f1)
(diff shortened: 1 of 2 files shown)


misc/scripts/add-repo.sh
```diff
@@ -22,47 +22,85 @@
...
 
-REPO="${2%/}"
-
-if [[ $REPO == *"github.com"* ]]; then
-    REPO="${REPO/'github.com'/'raw.githubusercontent.com'}"
-    if [[ $REPO != *"/tree/"* ]]; then
-        REPO="$REPO/master"
-        fancy_message warn "Assuming that git branch is ${GREEN}master${NC}"
+parse_repo() {
+    IFS=':' read -ra ADDR <<< "$1"
```
Elsie19, line 26: Localize `ADDR`.

Elsie19 (changes requested):

misc/scripts/add-repo.sh
```diff
@@ -22,47 +22,86 @@
...
+        if [[ $REPO != *"/tree/"* ]]; then
+            REPO="$REPO/-/raw/master"
+            fancy_message warn "Assuming that git branch is ${GREEN}master${NC}"
+        else
+            REPO="${REPO/"/tree/"/"/raw/"}"
+        fi
+        ;;
+    *"github:"*)
+        parse_repo "${REPO}"
+        REPO="https://raw.${PROV}usercontent.com/${USER}/${HEAD}/${BRANCH}"
```
oklopfer, line 61: this could potetionally be 
```bash
REPO="https://${PROV}.com/${USER}/${HEAD}/raw/${BRANCH}"
```
instead to look more equivalent to the gitlab one, but it just redirects to the usercontent link in the end anyways.

oklopfer: probably not worth having mixed and matched GitHub links

Elsie19 (approved):

PR #1129, opened 2024-05-27
chore: replace DT CoC with Contributor Covenant

Elsie19:
## Checklist

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,52 +1,64 @@
-<!--TITLE: The FOSS Code of Conduct
-AUTHOR: Derek Taylor
-TRANSLATED TO MARKDOWN BY: Naruto Uzumaki (NarutoXY)
-DATE: April 23, 2021-->
+# Contributor Covenant Code of Conduct
 
-# THE FOSS CODE OF CONDUCT
+## Our Pledge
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -1,52 +1,64 @@
...
+
+### 4. Permanent Ban
+Community Impact: Demonstrating a pattern of violation of community standards, including sustained inappropriate behavior, harassment of an individual, or aggression toward or disparagement of classes of individuals.
+
+Consequence: A permanent ban from any sort of public interaction within the community.
+
+## Coding Standards & Political Activism
+* "AI" tools are becoming more and more popular, and in order to make this community's code as high quality as possible, we forbid anyone from using AI tooling to produce code that you intend for others to use. You may use AI tooling for help, guidance and even debugging, but replacing actual programming work is strictly forbidden. As a rule of thumb, if an AI tool produces code that you wouldn't have been able to write on your own, you are forbidden from pushing that. If AI generated code is found in an open PR, you will be warned and told to remake that portion, and if AI generated code is merged into the codebase and found, whoever was responsible for that code will be immediately banned from the project.
```
oklopfer, line 62: rather than banning them, can we do something like stripping them of their maintainership? I don't think they should be extra punished for our own failure to recognize the generated code, but I think removing their maintainership is a fair and equal response.

vigress8:
We should quit stalling this; my stance remains that the Coding Standards & Political Activism section is unnecessary, and it's best to adopt the unmodified Contributor Covenant.

PR #1137, opened 2024-06-05
fix({build,srcinfo}.sh): missed stuff

oklopfer:
## Purpose

forgor to add all alt in one place and srcinfo had a bug with pacdeps

## Approach

fixem

## Progress

- [x] do it
- [x] test it

## Checklist


misc/scripts/srcinfo.sh
```diff
@@ -182,7 +182,7 @@ function srcinfo.parse() {
         fi
         mapfile -t split_up <<< "${part_two/_array_/$'\n'}"
         declare -n addarr="${split_up[0]}"
-
+        unset "${split_up[1]}"
```
Elsie19, line 185: Make it so upstream when you're done.

Elsie19 (approved):

PR #1142, opened 2024-06-08
refactor(pacstall): add has_command(), improve is_array()

vigress8:


pacstall
```diff
@@ -304,16 +304,22 @@ function check_url() {
     fi
 }
 
+# Checks if a command is available.
+# Plain "command -v" succeeds if the argument is in PATH, even if not executable.
+function has() {
```
oklopfer, line 309: can we make this `has_command` instead

pacstall
```diff
@@ -428,14 +434,7 @@ function is_apt_package_installed() {
...
 function is_array() {
-    local input raw
-    input="${1}"
-    raw="$(declare -p "${input}" 2> /dev/null)"
-    if [[ ${raw} == "declare -a ${input}"* ]]; then
-        return 0
-    else
-        return 1
-    fi
+    [[ ${!1@a} == *a* ]]
```
oklopfer, line 437: we should use this in https://github.com/pacstall/pacstall/blob/931763db64584a2798eaaecdd6d86aa1f2d7455b/misc/scripts/srcinfo.sh#L188 +
https://github.com/Elsie19/srcinfo_bash/blob/649141624f34844c280711de4a41c6a4be0f0faf/src/srcinfo.sh#L164 too

oklopfer (changes requested):

pacstall
```diff
@@ -953,11 +952,6 @@ Helpful links:
                 exit 1
             }
 
-            if curl --output /dev/null --silent --head --fail https://raw.githubusercontent.com/"$USERNAME"/pacstall/"$BRANCH"/pyproject.toml || curl --output /dev/null --silent --head --fail https://raw.githubusercontent.com/"$USERNAME"/pacstall/"$BRANCH"/Cargo.toml; then
```
oklopfer, line 956: This section is still required. Even though no upstream branch exists for these, if someone is working on pacstall-py or pacstall-rs on a repo called simply pacstall, this prevents an upgrade to that repo.

oklopfer (changes requested):

pacstall
vigress8: A rewrite is not even happening currently. If we're gonna cover entirely hypothetical situations, might as well add this check for every programming language. Makes no sense and claiming such a check is "required" is absurd

oklopfer: https://github.com/pacstall/pacstall/pull/460 it's been there for years. there's no reason to be removing it now. there equally makes no sense for you not to just leave it alone.

vigress8: It's at least one redundant curl query, which makes for worse performance. That's why I removed it

Elsie19: It's not redundant. It's possible in the future we might have a rewrite still, I say we should not remove this.

oklopfer (approved):

Elsie19 (approved):

PR #1155, opened 2024-06-22
refactor(*): parse parent distro if exists

oklopfer:
## Purpose
for distros like pop, elementary, mint, etc, they do not get special enhanced arrays. also removing lsb-release dependency

## Approach
allow them to use the parent distro for enhanced arrays and compatibility checks. This works in different ways for each piece:
#### `set_distro`
the following distro examples are described with the combination `child:codename/num|parent:codename/num`:
```bash
debian:trixie/13|debian:sid/sid
elementary:next/8|ubuntu:noble/24.04
linuxmint:elsie/5|debian:bullseye/11
linuxmint:virginia/21.3|ubuntu:jammy/22.04
pop:focal/20.04|ubuntu:focal/20.04
ubuntu:devel/2024.1
ubuntu:oracular/24.10
```
running `set_distro` results in:
```bash
debian:trixie
elementary:next
linuxmint:elsie
linuxmint:virginia
pop:focal
ubuntu:devel
ubuntu:oracular
```
and `set_distro parent` results in:
```bash
debian:sid
ubuntu:noble
debian:bullseye
ubuntu:jammy
ubuntu:focal
ubuntu:devel
ubuntu:oracular
```
enhanced arrays use the `DISTRO` variable, so `DISTRO` is now equal to `set_distro parent`. We have a new variable for pacscript usage `CDISTRO`, which will be the true current distro, as the `DISTRO` variable previously worked.
#### `get_compatible_releases`
In all cases, if *either* the parent or true distro is found to match in the `compatible` array, then it is marked as compatible. This means in an example where distro is `linuxmint:virginia`, if the compatible array lists `ubuntu:jammy` but not `linuxmint:virginia`, it is still recognized as compatible. For `linuxmint:virginia`, the following would all trigger compatibility:
```bash
"linuxmint:virginia"
"linuxmint:21.3"
"linuxmint:*"
"*:virginia"
"*:21.3"
"ubuntu:jammy"
"ubuntu:22.04"
"ubuntu:*"
"*:jammy"
"*:22.04"
```
#### `get_incompatible_releases`
If no trace of true distro is found in the `incompatible` array:
```bash
"${distro_name}:${distro_version_name}"
"${distro_name}:${distro_version_number}"
"${distro_name}:*"
"*:${distro_version_name}"
"*:${distro_version_number}"
```
then if parent distro exists, and if the `incompatible` array contains either `"*:${distro_parent_vname}"` or `"*:${distro_parent_number}"`, override our incompatible check variables to:
```bash
distro_name="${distro_parent}"
distro_version_name="${distro_parent_vname}"
if [[ -n ${distro_parent_number} ]]; then
    distro_version_number="${distro_parent_number}"
fi
```
this means for `linuxmint:virginia`, the following would trigger incompatibility:
```bash
"linuxmint:virginia"
"linuxmint:21.3"
"linuxmint:*"
"*:virginia"
"*:21.3"
"*:jammy"
"*:22.04"
```
the following would *not* trigger incompatibility:
```bash
"ubuntu:*"
"ubuntu:jammy"
"ubuntu:22.04"
```
## Progress

- [x] do it 
- [x] test it

## Checklist

### DIFF
--- a/misc/scripts/fetch-sources.sh
+++ b/misc/scripts/fetch-sources.sh
@@ -540,11 +540,19 @@ function get_incompatible_releases() {
     # example for this function is "ubuntu:jammy"
     local distro_name distro_version_name distro_version_number distro_parent distro_parent_vname distro_parent_number incomp_list=("${@,,}")
     calc_distro
-    if ! array.contains incomp_list "${distro_name}:${distro_version_name}"; then
-        if [[ -n ${distro_parent_vname} ]] && array.contains incomp_list "${distro_parent}:${distro_parent_vname}"; then
+
+
+    if ! array.contains incomp_list "${distro_name}:${distro_version_name}" && \
+        ! array.contains incomp_list "${distro_name}:${distro_version_number}" && \
+        ! array.contains incomp_list "${distro_name}:*" && \
+        ! array.contains incomp_list "*:${distro_version_name}" && \
+        ! array.contains incomp_list "*:${distro_version_number}"; then
+        if [[ -n ${distro_parent_vname} ]]; then
             distro_name="${distro_parent}"
-            distro_version_number="${distro_parent_number}"
             distro_version_name="${distro_parent_vname}"
+            if [[ -n ${distro_parent_number} ]]; then
+                distro_version_number="${distro_parent_number}"
+            fi
         fi
     fi
     for key in "${incomp_list[@]}"; do
--- a/misc/scripts/fetch-sources.sh
+++ b/misc/scripts/fetch-sources.sh
@@ -482,7 +482,7 @@ function calc_distro() {
             # have to set this empty instead of unsetting as the local is higher up
             distro_parent_vname=""
         else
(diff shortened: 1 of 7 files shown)


misc/scripts/fetch-sources.sh
```diff
@@ -459,64 +459,106 @@ function append_modifier_entries() {
...
 function calc_distro() {
-    distro_name="$(lsb_release -si 2> /dev/null)"
-    distro_name="${distro_name,,}"
-    if [[ "$(lsb_release -ds 2> /dev/null | tail -c 4)" == "sid" ]]; then
-        distro_version_name="sid"
-        distro_version_number="sid"
-    else
-        distro_version_name="$(lsb_release -sc 2> /dev/null)"
-        distro_version_number="$(lsb_release -sr 2> /dev/null)"
+    local distro_pretty_name
```
Elsie19, line 462:
```suggestion
    local distro_pretty_name key value
```

Elsie19 (changes requested):

Elsie19 (approved):

PR #1156, opened 2024-06-24
refactor(*): `stacktrace` and safe `rm`s

oklopfer:
## Purpose

set -x is so hard to read sometimes

## Approach

trap everything, trace it back
- All files and functions must have `{ ignore_stack=false; set -o pipefail; trap stacktrace ERR RETURN; }` at the top
- Any time `return 1` is provided by an internal function, it should be immediately preceded by `ignore_stack=true`, so that intentional error outs are not caught by the trace
- `stacktrace` will pipe back out the return codes it receives that are not `1`

in fixing this PR, three other changes/fixes happened:
1. if `depends` attributes have no aptitude match (like optdepends), then the script will fail out
    - this also fixed optdepends from checking with apt-cache to checking with aptitude for arch enabled depends
    - this also fixed all of the aptitude checks to properly check for architecture
2. `-git` packages now have a proper version check
    - the git commit is split from the rest of the `pkgver`, and considered separately
    - first, if the git commits of the local and remote version do not match, then an update is triggered regardless of the rest of the version
    - if the git commits are the same, then the normal version is considered (`${pkgver}-pacstall${pkgrel}`)
3. relative files will check if they exist locally relative to the pacscript first, before defaulting to downloading like normal

## Progress

- [x] do it
- [x] test it
- [x] make it stack everything
- [x] make sure things all work normally and don't falsely stacktrace

Screenshot 2024-06-27 at 6 21 39 PM

## Checklist

### DIFF
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -25,6 +25,7 @@
 { ignore_stack=false; set -o pipefail; trap stacktrace ERR; }
 
 function parse_repo() {
+    # shellcheck disable=SC2034
     { ignore_stack=false; set -o pipefail; trap stacktrace ERR; }
     local ADDR
     IFS=':' read -ra ADDR <<< "$1"
--- a/misc/scripts/add-repo.sh
+++ b/misc/scripts/add-repo.sh
@@ -22,10 +22,10 @@
 # You should have received a copy of the GNU General Public License
 # along with Pacstall. If not, see <https://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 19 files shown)


oklopfer:
I think dep tree needs patching

oklopfer:
optdeps also does not pre-find packages labeled with :arch

oklopfer:
Screenshot 2024-06-28 at 2 20 52 PM

uh oh

Elsie19 (approved):

PR #1195, opened 2024-07-22
feat(*): add regular `suggests` array

oklopfer:
## Purpose

because of how optdeps works, sometimes when building static packages one needs to just use the suggests array.

## Approach

refactor `prompt_optdepends`

## Progress

- [x] do it
- [x] test it (I am doubtful it will work right away)

## Checklist

### DIFF
--- a/misc/scripts/build.sh
+++ b/misc/scripts/build.sh
@@ -217,6 +217,7 @@ function prompt_optdepends() {
 
 function deblog_depends() {
     { ignore_stack=false; set -o pipefail; trap stacktrace ERR RETURN; }
+    # shellcheck disable=SC2034
     local log_depends log_depends_str input_depends="${1}" todeblog="${2}"
     dep_const.format_control "${input_depends}" log_depends
     dep_const.comma_array log_depends log_depends_str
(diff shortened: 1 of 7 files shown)


oklopfer:
seems to work right away

oklopfer:
Will try to fix #1196 while im here

oklopfer:
nvm I'll solve that in a new pr, it is unrelated

misc/scripts/build.sh
```diff
@@ -183,90 +254,11 @@ function prompt_optdepends() {
...
-            (
-                #shellcheck disable=SC1090
-                source "$METADIR/$i"
-                if [[ -n $_gives ]]; then
-                    echo "$_gives" | tee -a "${PACDIR}-gives-${pacname}" > /dev/null
-                else
-                    echo "$_name" | tee -a "${PACDIR}-gives-${pacname}" > /dev/null
-                fi
-            )
+            awk -F'=' '/^_gives=/{gives=$2} /^_name=/{name=$2} END{print (gives ? gives : name)}' "${METADIR}/${i}" >> "${PACDIR}-gives-${pacname}"
```
Elsie19, line 261: If this is actually easier then fine, but I'm just saying I don't like the amount of awk creeping into the codebase.

Elsie19 (approved):

PR #1229, opened 2024-08-29
fix({build,fetch-sources,version-constraints}.sh): `all` checks

oklopfer:
## Purpose

if a package has both version constraints and is only available as `all`, it needs that as its architecture upon search

## Approach

do that check and also clean up the other aptitude checks

## Progress

- [x] do it
- [x] test it

## Addendum

https://github.com/pacstall/pacstall-programs/pull/6368

## Checklist

### DIFF
--- a/misc/scripts/build.sh
+++ b/misc/scripts/build.sh
@@ -56,6 +56,30 @@ function clean_builddir() {
     sudo rm -f "${STAGEDIR:?}/${pacname}.deb"
 }
 
+function check_gen_dep() {
+    { ignore_stack=false; set -o pipefail; trap stacktrace ERR RETURN; }
+    local onlyname="${1}" onlyarch="${2}" onlyreal="${3}" onlywhere="${4}"
+    if [[ ${onlyname} == *":${onlyarch}" ]]; then
+        if [[ -z "$(aptitude search --quiet --disable-columns "?exact-name(${onlyname%:*})?architecture(${onlyarch})" -F "%p")" ]]; then
+            if [[ -z "$(aptitude search --quiet --disable-columns "?provides(^${onlyname%:*}$)?architecture(${onlyarch})" -F "%p")" ]]; then
+                fancy_message sub $"%b [required]" "${CYAN}${onlyreal}${NC} ${RED}✗${NC}"
+                echo "${onlyreal}" >> "${onlywhere}"
+                return 0
+            fi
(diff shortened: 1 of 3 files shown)


misc/scripts/build.sh
```diff
@@ -56,6 +56,38 @@ function clean_builddir() {
...
+                        if [[ -z "$(aptitude search --quiet --disable-columns "?provides(^${onlyname}$)?architecture(all)" -F "%p")" ]]; then
+                            fancy_message sub $"%b [required]" "${CYAN}${onlyreal}${NC} ${RED}✗${NC}"
+                            echo "${onlyreal}" >> "${onlywhere}"
+                            return 0
+                        fi
+                    fi
+                fi
+            fi
+        fi
+    fi
```
vigress8, line 88:
```suggestion
    if [[ ${onlyname} == *":${onlyarch}" &&
        -z "$(aptitude search --quiet --disable-columns "?exact-name(${onlyname%:*})?architecture(${onlyarch})" -F "%p")" &&
        -z "$(aptitude search --quiet --disable-columns "?provides(^${onlyname%:*}$)?architecture(${onlyarch})" -F "%p")" ]] \
        || [[ -z "$(apt-cache search --no-generate --names-only "^${onlyname}\$" 2> /dev/null || apt-cache search --names-only "^${onlyname}\$")" &&
        -z "$(aptitude search --quiet --disable-columns "?exact-name(${onlyname})?architecture(${onlyarch})" -F "%p")" &&
        -z "$(aptitude search --quiet --disable-columns "?provides(^${onlyname}$)?architecture(${onlyarch})" -F "%p")" ]]; then
        fancy_message sub $"%b [required]" "${CYAN}${onlyreal}${NC} ${RED}✗${NC}"
        echo "${onlyreal}" >> "${onlywhere}"
        return 0
    fi

    return 1
```

oklopfer: Not sure why the else was removed, this breaks the logic

vigress8: Is the or-statement not equivalent to the if-else?
Both branches are doing the exact same thing

oklopfer: I see it now the suggestion format just was impossible to read

it should not have the return 1 tho

vigress8: Updated

PR #1189, opened 2024-07-12
feat(repo): add aliases and specifiers

Elsie19:
## Purpose

Currently our `pacstallrepo` file only supports raw URLs per newline, which really isn't great. Now we can add aliases to quickly reference those repos, and specifiers to narrow down ambiguous repository names.

## Approach

`pacstallrepo` now optionally supports aliases, specifiers, or both. Here's an example:
```bash
[github] https://raw.githubusercontent.com/pacstall/pacstall-programs/master @pacstall
```

## Progress

- [x] Create library file.
- [x] Create manpages.
- [ ] Integrate it.

## Checklist

- [x] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

PR #1239, opened 2024-09-04
fix(bash): add zsh exception for mapfile

Elsie19:
## Purpose

I goofed on #1189 at the part about getting the urls from `pacstallrepo`. ZSH should no longer scream mapfile errors.

## Checklist

### DIFF
--- a/misc/completion/bash
+++ b/misc/completion/bash
@@ -25,6 +25,8 @@ _pacstall_get_repo_urls() {
     local part split
     while read -r part; do
         if [[ -n $ZSH_NAME ]]; then
+            # shellcheck disable=SC2296
+            # shellcheck disable=SC2116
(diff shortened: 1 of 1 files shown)


misc/completion/bash
```diff
@@ -24,7 +24,13 @@
 _pacstall_get_repo_urls() {
     local part split
     while read -r part; do
-        mapfile -t split <<< "${part// /$'\n'}"
+        if [[ -n $ZSH_NAME ]]; then
+            # shellcheck disable=SC2296
+            # shellcheck disable=SC2116
```
oklopfer, line 29:
```suggestion
            # shellcheck disable=SC2296,SC2116
```

oklopfer (approved):

PR #1250, opened 2024-09-20
previous command crashes terminal when run under zsh shell

sanjay-kr-commit:
## Purpose

to avoid terminal crash when run under zsh shell

## Approach

it runs current command under a bash shell to avoid zsh shell crash

## Progress

- [x] Make command compatible with zsh shell

## Checklist

- [ ] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,14 +38,14 @@
 You can run the command below to install Pacstall.
 You can also grab the deb file [here](https://github.com/pacstall/pacstall/releases/latest) but it may be a bit older.
 ```bash
-sudo bash -c "$(curl -fsSL https://pacstall.dev/q/install || wget -q https://pacstall.dev/q/install -O -)"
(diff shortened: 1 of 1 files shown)


Elsie19:
Can you explain what "crashes"?

sanjay-kr-commit:
> Can you explain what "crashes"?

Today I did a clean install of my distro but whenever I tried pasting previous command in zsh shell it kept crashing and worked under bash for some reason, although I'm not sure why, so I wrapped the command under another bash instance and it worked

Elsie19:
You're not explaining what a crash means. Did it segfault? Did it just fail to run?

sanjay-kr-commit:
> You're not explaining what a crash means. Did it segfault? Did it just fail to run?

Issue #1243, opened 2024-09-10
BUG: provides search can result in multiple pkgver responses


oklopfer:
### Describe the bug

```
[version-constraints.sh:dep_const.apt_compare_to_constraints():47] - DEBUG: aptitude search --quiet --disable-columns '?provides(^java-runtime$)?architecture(amd64)' -F %V
[version-constraints.sh:dep_const.apt_compare_to_constraints():47] - DEBUG: pkg_version='2:1.21-76
11.0.24+8-1ubuntu3
17.0.12+7-1ubuntu2
21.0.4+7-3
22.0.2+9-4
23~37ea-1
24~14ea-1
8u422-b05-1ubuntu1'
[version-constraints.sh:dep_const.apt_compare_to_constraints():50] - DEBUG: case "${compare_pkg}" in
[version-constraints.sh:dep_const.apt_compare_to_constraints():53] - DEBUG: dpkg --compare-versions '2:1.21-76
11.0.24+8-1ubuntu3
17.0.12+7-1ubuntu2
21.0.4+7-3
22.0.2+9-4
23~37ea-1
24~14ea-1
8u422-b05-1ubuntu1' ge 21
dpkg: warning: version '2:1.21-76
11.0.24+8-1ubuntu3
17.0.12+7-1ubuntu2
21.0.4+7-3
22.0.2+9-4
23~37ea-1
24~14ea-1
8u422-b05-1ubuntu1' has bad syntax: invalid character in version number
[version-constraints.sh:dep_const.apt_compare_to_constraints():53] - DEBUG: ret=0
```

### To reproduce

asked for `optdepends=("java-runtime>=21: blah")`

### Expected behavior

probably should just pick first (top) result

### Screenshots

<img width="541" alt="Screenshot 2024-09-10 at 5 51 35 PM" src="https://github.com/user-attachments/assets/aa980914-06f2-496b-ba4d-6048d2b93f20">

### OS version

Rhino Linux

### Pacstall version

5.4.1

PR #1252, opened 2024-09-22
fix(version-constraints.sh): loop over multiple provides

oklopfer:
## Purpose

see https://github.com/pacstall/pacstall/issues/1243

## Approach

loop over them, as long as at least one of the results satisfies our check, we don't care, apt can take it from there

## Progress

- [x] do it
- [x] test it

## Addendum

https://github.com/pacstall/pacstall/issues/1243

## Checklist


oklopfer:
can't release 5.5.0 till this is reviewed

vigress8:
Looks fine to me, but need a test case

Elsie19 (approved):
lGTM

PR #1271, opened 2024-11-17
feat({fetch-sources,package}.sh): allow extract-to syntax for `source`

oklopfer:
## Purpose

Allows us to git rid of/work around `_archive` by explicitly defining the name of the extracted folder

NOTE: `_archive` will be removed entirely in a separate PR

## Approach

- Where before we allowed:
    - `custom-location.zip::git+file://example.com/archive.tar.xz?h=file#branch=master`
- We now allow (in addition):
    - `custom-location.zip@extracted_location::git+file://example.com/archive.tar.xz?h=file#branch=master`
    - `@extracted_location::git+file://example.com/archive.tar.xz?h=file#branch=master`

 Where one can then use `cd extracted_location` instead of `cd ${_archive}` (the variable is internally called `${to_location}`, per source)

For `.git` and non-extractable sources, `${dest}` already performed this function, so `${to_location}` is ignored.

## Progress

- [x] do it
- [x] test it basically
- [x] test it extensively
- [x] figure out how to ensure no conflicts with `dest` and `to_location`

## Checklist

### DIFF
--- a/misc/scripts/fetch-sources.sh
+++ b/misc/scripts/fetch-sources.sh
@@ -32,11 +32,15 @@ source "${SCRIPTDIR}/scripts/build.sh" || {
 
 function parse_source_entry() {
     { ignore_stack=false; set -o pipefail; trap stacktrace ERR RETURN; }
-    unset source_url dest git_branch git_tag git_commit
+    unset source_url dest git_branch git_tag git_commit to_location
     local entry="$1"
     source_url="${entry#*::}"
     dest="${entry%%::*}"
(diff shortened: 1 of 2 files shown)


oklopfer:
To test this one:
1. find a package using `${_archive}`
2. change the `${source[0]}` entry to be `"@${somevariable}::${source[0]}"`
3. change all instances of `${_archive}` to `${somevariable}`

oklopfer:
def has some bugs

PR #1274, opened 2024-12-01
ref(pacstall)!: do not allow update if dpkg installed

oklopfer:
## Purpose

breaking out the .deb from the curl install

## Approach

dont allow update if pacstall is installed from the deb

## Progress

- [x] do it
- [x] test it

## Checklist

- [x] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

PR #1275, opened 2024-12-06
fix(suggested_solution): use double dash as separator instead of "ZZZZ"

Elsie19:
## Purpose

It looks a lot nicer and is more immediately recognized as a separator.

## Other

Should be merged after #1274 because it touches the same areas.

## Checklist


oklopfer:
This can be rebased now to include the new message

Elsie19:
@oklopfer rereview.

oklopfer (approved):
works lgtm

PR #1280, opened 2024-12-19
feat: broken package checking

Elsie19:
## Purpose

Because of certain packages apparently borking multiple systems (`wine-staging-git`), I want to implement a system of guards where we can simply drop an entry into a file and declare a package quarantined, where users are warned of problems while still being able to install it if they wish.

## Approach

The entries go in `broken-packages` in the repo root, and every pacscript download, it will check the file.

## Addendum

https://github.com/pacstall/pacstall-programs/pull/6763
https://github.com/pacstall/pacstall-programs/issues/6357

## Checklist

### DIFF
--- a/misc/scripts/get-pacscript.sh
+++ b/misc/scripts/get-pacscript.sh
@@ -26,6 +26,36 @@
 
 { ignore_stack=false; set -o pipefail; trap stacktrace ERR RETURN; }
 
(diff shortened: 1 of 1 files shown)


Elsie19:
True, because then we could check for pkgbase/pkgchild stuff like that as well 🤔

PR #1283, opened 2024-12-23
feat(package): show reinstall warning

### LINKED ISSUE #1282: Do not reinstall "installed" packages with "pacstall -I"
### Is your feature related to a problem? Please describe.

It seems like a good idea to avoid reinstalling "installed" packages.

### A clear and concise description of what you want to happen.

I want to install a package (for example: discord)
```sh
pacstall -PI discord-deb
```

but this package is already installed so pacstall prints to stdout something like this:
```
[discord] is already installed at version [VERSION]
```

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

Elsie19:
## Purpose

Fixes #1282.

## Approach

Instead of telling the user it's already installed, show a warning.

## Addendum

@oklopfer idk if pacname and pkgver still refer to what they should so I probably might have messed those up.

## Checklist

### DIFF
--- a/misc/scripts/package.sh
+++ b/misc/scripts/package.sh
@@ -53,6 +53,10 @@ if ((PACSTALL_INSTALL == 0)) && [[ ${pacname} == *-deb ]]; then
     return 0
 fi
 
+if is_package_installed "${pacname}" && dpkg --compare-versions "${pkgver}" eq "$(dpkg-query --showformat='${Version}' --show "${pacname}")"; then
+    fancy_message warn $"Reinstalling '%s'" "${pacname}"
+fi
+
 masked_packages=()
 getMasks masked_packages
 if ((${#masked_packages[@]} != 0)); then
--- a/misc/scripts/package.sh
+++ b/misc/scripts/package.sh
@@ -53,10 +53,6 @@ if ((PACSTALL_INSTALL == 0)) && [[ ${pacname} == *-deb ]]; then
     return 0
 fi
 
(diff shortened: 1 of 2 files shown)


misc/scripts/package.sh
```diff
@@ -53,6 +53,10 @@ if ((PACSTALL_INSTALL == 0)) && [[ ${pacname} == *-deb ]]; then
     return 0
 fi
 
+if is_package_installed "${gives:-${pacname}}" && dpkg --compare-versions "${pkgver}" eq "$(dpkg-query --showformat='${Version}' --show "${gives:-${pacname}}")"; then
```
oklopfer, line 56: just realized this doesnt work bc dpkg version is equal to full_version and not pkgver

oklopfer: spectacular news though looks like I set it a little down in the file so maybe just run this check after that instead?

https://github.com/pacstall/pacstall/blob/c2fbf25b6693e68d8dbc0eb5a2b0cd16ba44d884/misc/scripts/package.sh#L115-L123

Elsie19: Done.

misc/scripts/package.sh
```diff
@@ -118,6 +118,10 @@ else
     full_version="${epoch+$epoch:}${pkgver}-pacstall${pkgrel:-1}"
 fi
 
+if is_package_installed "${pacname}" && dpkg --compare-versions "${full_version}" eq "$(dpkg-query --showformat='${Version}' --show "${pacname}")"; then
```
Elsie19, line 121:
```suggestion
if is_package_installed "${pacname}" && dpkg --compare-versions "${full_version}" eq "$(dpkg-query --showformat='${Version}' --show "${gives:-${pacname}}")"; then
```

oklopfer (approved):

PR #1314, opened 2025-01-17
ci: use arm64 runners for docker

oklopfer:
## Purpose

pros: will run docker builds wayyyyyy faster

## Approach

cons: have to explicitly define each matrix combo now

## Checklist

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -21,10 +21,50 @@ jobs:
   build:
     strategy:
       matrix:
-        OS: ["ubuntu:latest", "ubuntu:devel", "ubuntu:rolling", "debian:stable", "debian:testing", "debian:unstable"]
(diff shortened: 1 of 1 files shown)


Elsie19:
Could this be extended to pacstall/pacstall-programs?

oklopfer:
> Could this be extended to pacstall/pacstall-programs?

Yes but I’m more cautious about adding those right now. 

Pros:
- `arm64` hashes and builds would always get checked properly
- `arm64`-only binaries would build faster
- CI’s reliance on `FARCH` could be dropped

Cons:
- Our runner count for each PR commit would double from 12 to 24, because in programs we run it on two branches
- There are currently no `arm64`-only binaries as far as I know, only debs, and `FARCH` works just fine to test those installs
- The CI would constantly be a Christmas tree for packages that support one architecture and not the other

PR #1334, opened 2025-02-23
feat(pacstall): allow env to define TMPDIR

oklopfer:
## Purpose

Some packages are too big for some device's tmpdirs

## Approach

give an envar for users to override the `/tmp` part of `PACDIR` (so if someone tries `PACSTALL_TMPDIR=/` it won't nuke them)

## Progress

- [x] do it
- [ ] test it

## Addendum

https://github.com/pacstall/pacstall/issues/1324

## Checklist

- [x] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

PR #1341, opened 2025-03-02
fix(*): respect TMPDIR override

oklopfer:
## Purpose

#1334 didn't do enough

## Approach

make the TMPDIR override its own variable that falls back to /tmp, and use that in place of /tmp anywhere in da code

## Progress

- [x] do it
- [x] test it

## Checklist

### DIFF
--- a/misc/scripts/update.sh
+++ b/misc/scripts/update.sh
@@ -33,13 +33,20 @@ export BPurple='\033[1;35m'
 METADIR="/var/lib/pacstall/metadata"
 LOGDIR="/var/log/pacstall/error_log"
 SCRIPTDIR="/usr/share/pacstall"
-PACDIR="/tmp/pacstall"
+PACTMP="${PACSTALL_TMPDIR:-/tmp}"
+PACDIR="${PACTMP}/pacstall"
 MAN8DIR="/usr/share/man/man8"
 MAN5DIR="/usr/share/man/man5"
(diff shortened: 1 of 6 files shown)


misc/scripts/update.sh
```diff
@@ -33,14 +33,21 @@ export BPurple='\033[1;35m'
...
+PACTMP="${PACSTALL_TMPDIR:-/tmp}"
+PACDIR="${PACTMP}/pacstall"
 MAN8DIR="/usr/share/man/man8"
 MAN5DIR="/usr/share/man/man5"
 PODIR="${SCRIPTDIR}/po"
 BASH_COMPLETION_DIR="/usr/share/bash-completion/completions"
 FISH_COMPLETION_DIR="/usr/share/fish/vendor_completions.d"
 PACSTALL_USER=$(logname 2> /dev/null || echo "${SUDO_USER:-${USER:-$(whoami)}}")
 
+if ! $(cd "${PACTMP}" 2> /dev/null); then
```
Elsie19, line 45: I know this is already merged but this is in a subshell, so it won't actually cd into it.

D-Brox: @Elsie19 That's kind of the point, no? It's just testing if it is accessible

oklopfer: Correct, we don’t want it to actually enter.

PR #1398, opened 2025-10-27
feat: sudo must be used in some options

aKqir24:
## Purpose
When trying to run the pacstall script without `sudo`, the script runs, however it calls `sudo` a bunch of times, also asking for the password many times, which is a bit redundant. I was thinking of making much simpler, by making the program exit with an error when not running it in `sudo` as shown in the picture below. Sorry for my poor English.

gscreenshot_2025-10-27-214342

## Approach

I added a `need_sudo` function:

```bash
function need_sudo() {
    if [ "$(id -u)" -ne 0 ]; then
        fancy_message error $"You cannot perform this operation unless you are root" >&2
        exit 1
    fi
}
```
which is called when a option that needs sudo is chosen.

## Progress

- [x] Add the `need_sudo` function to all options
- [x] Make sure the other scripts uses the `sudo` in options
- [x] Verify if the options really uses `sudo`
- [ ] Test every options in some conditions

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -1369,6 +1369,8 @@ Helpful links:
                 PRNUM="${BASH_REMATCH[3]}"
                 METAURL="${BASH_REMATCH[5]}"
             fi
+            need_sudo
+
             # shellcheck source=./misc/scripts/quality-assurance.sh
             source "$SCRIPTDIR/scripts/quality-assurance.sh"
             exit 0


oklopfer:
Apologies for not having responded to this sooner. Regardless, thank you for making contributions to Pacstall!

aKqir24:
> Apologies for not having responded to this sooner. Regardless, thank you for making contributions to Pacstall!

Additionally, I’d like to point out an unusual behavior I found when pacstall is executed without root privileges, it repeatedly prompts for the sudo password. Is this normal?

D-Brox:
That's probably related to the sudo setting `timestamp_timeout`, I think the default is 5 minutes

PR #1408, opened 2025-11-21
fix(pacstall): Sort files for tree display

tdryer:
`dpkg --listfiles` prints files in an unspecified order.

For example:

    $ dpkg --listfiles less | grep -C 1 /usr/bin/lessfile
    /usr/share/man/man1/lesspipe.1.gz
    /usr/bin/lessfile
    /usr/share/man/man1/lessfile.1.gz

Sort the files used for `pacstall --tree <package>` to ensure the tree is drawn correctly.

See https://github.com/pacstall/pacstall-programs/pull/8213.

- [X] I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -1307,7 +1307,7 @@ Helpful links:
 
             source "$METADIR/$PACKAGE"
 
-            dpkg --listfiles "${_gives:-$_name}" | sed -e "s/[^-][^\/]*\//  │/g" -e "s/│\([^ ]\)/│──\1/"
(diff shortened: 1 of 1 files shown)


Elsie19:
https://salsa.debian.org/dpkg-team/dpkg/-/blob/main/src/query/main.c#L505-L571

This is where the code for `-L` is defined.

PR #1412, opened 2025-12-03
feat(main): only require elevating once

### LINKED ISSUE #1236: Handle SIGINT/^C and sudo in a way that doesn't ask for sudo after SIGINT
### Is your feature related to a problem? Please describe.

I really have a problem with how sudo is called in pacstall.  Under no circumstances should/do I trust a program I've _tried_ to ^C out of with my sudo password. 

```
[*] WARNING: 
Interrupted, cleaning up
[sudo] password for $USER:
sudo: a password is required
```

You also can use sudo -p here to explain WHY I need to enter my sudo password after I've sent SIGINT to the program to end it.  

Anyway, if I want to get out of it [p]kill are needed.  As it happens the lowest level that pacstall can be exited from with pkill is -4, SIGILL, which I find appropriate because I already sent SIGINT and I feel that asking for the admin password after I've asked to get out of the program is, in fact, faulty and unknown: `SIGILL - When a process performs a faulty, forbidden, or unknown function, the system sends the SIGILL signal to the process. This is the “ILLegal SIGnal”.` 

Functionally, I doubt it's much different from [p]kill -9 but i've been trying not to default to [p]kill -9 from the start anymore.

### A clear and concise description of what you want to happen.

apt doesn't do this.  i can't think of anything else that demands my sudo password to get out of it after receiving SIGINT.  I guess if there are other programs I've run across that do this, I do not end up using them.  Pacstall is somewhat essential in maintaining Rhino so I've had few good choices but but to grin and bear it.

if pacstall needs sudo, why can't you just write it so that it must be called as `sudo pacstall ...`?

if it **must** be called from in the program then it should both be explained with `sudo -p` so that the user understands what the sudo password is being used for and it should be expired at the end of the script with sudo -k.

### Describe alternatives you've considered

_No response_

### Additional context

If pacstall follows best practices for sudo and everything else does not then please point me to a reference which shows that this is the right way to handle sudo as opposed to the way e.g., apt handles it, otherwise, I really find it poor practice to get users in the habit of giving sudo this way.

D-Brox:
## Purpose

Reducing the number of times sudo password is required.
Fixing bugs related to sudo-rs timeout.

## Approach

Elevate user when required.
Make lock only check when elevated.
Apt upgrade only when needed.

## Progress

- [x] Do it
- [x] Test it

## Addendum

Another approach to what #1398 intended to do.

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -1316,8 +1316,7 @@ Helpful links:
                 exit 1
             fi
 
-            elevate
-            apt_update
+            [[ $LIST_ONLY ]] || { elevate; apt_update }
 
             check_url "https://github.com" 2> /dev/null || check_url "https://gitlab.com" 2> /dev/null || {
                 fancy_message error $"Could not connect to the internet"
--- a/pacstall
+++ b/pacstall
@@ -801,14 +801,15 @@ function elevate() {
         ${PACSTALL_VERBOSE} || quiet="-Q"
         [[ $NOSANDBOX ]] && nosandbox="-Ns"
 
+        trap - ERR RETURN
+        unset ignore_stack
         sudo PACSTALL_SUPPRESS_SOLUTIONS="$PACSTALL_SUPPRESS_SOLUTIONS" \
             PACSTALL_BUILD_CORES="$PACSTALL_BUILD_CORES" PACSTALL_EDITOR="$PACSTALL_EDITOR" \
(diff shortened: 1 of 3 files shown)


pacstall
```diff
@@ -1266,6 +1299,9 @@ Helpful links:
                 exit 1
             fi
 
+            elevate
```
oklopfer: Problematic placement

oklopfer (changes requested):

Elsie19 (changes requested):
Update the `.pot` file.

oklopfer (approved):

PR #1433, opened 2026-03-03
fix({checks,install,update}.sh): spdx got another update

oklopfer:
## Purpose
sobs

## Approach
update spdx version dep, and use whatever licenses are on system (if you mess with them that's your issue)

## Progress

- [x] do it
- [x] test it

## Checklist


misc/scripts/checks.sh
```diff
@@ -659,7 +659,7 @@ function lint_priority() {
...
+    local ret=0 linlicense idx=0 license_list=($(printf '%s\n' /usr/share/spdx-licenses/xml/*.xml | rev | cut -d. -f2- | cut -d/ -f1 | rev))
```
Elsie19:
```suggestion
    local ret=0 linlicense idx=0 license_list=($(ls /usr/share/spdx-licenses/xml | xargs -I'{}' basename '{}' .xml))
```

Elsie19 (changes requested):

Elsie19 (approved):

PR #1451, opened 2026-03-26
fix(bwrap): declare arrays in bwrapenv

### LINKED ISSUE #1449: Bash arrays not propagated into sandboxed invocation of functions
# Description

When a variable is defined in the body of a script, the definition is available within a function, invoked from the sandboxed environment, only if the variable is scalar, rather than an indexed or associative array.

# Test case

The following example file will be saved as `pactest.pacscript`:

```
pkgname=pactest
pkgdesc="Test case"
license=("Apache-2.0")
pkgver="1.0.0"
source=("pactest.tar")
arch=("any")
maintainer=("Jon Do <jondo@example.none>")

scalar=foo

list=(
  bar
  baz
)

declare -A dict=(
  [start]=dawn
  [end]=dusk
)

package() {

  echo $scalar

  for val in ${list[@]}; do
    echo $val;
  done

  echo ${dict[start]}
  echo ${dict[end]}
  
}
```

The following behavior can be confirmed, with Bash 5.2.21:

```
$ ( cat pactest.pacscript ; echo package ) | bash
foo
bar
baz
dawn
dusk
```

In contrast, with Pacstall 6.3.11, the following also may be confirmed:

```
$ sudo pacstall -PI pactest.pacscript 
[*] WARNING: Prompts are disabled
(pactest) Do you want to view/edit the pacscript? [y/N] N
[+] INFO: Sourcing pacscript
[+] INFO: Checking build dependencies
	[>] tar ✓ [installed]
[+] INFO: Checking apt dependencies
[+] INFO: Retrieving packages
[+] INFO: Copying local archive pactest.tar
	[>] Extracting pactest.tar
[+] INFO: Running functions
	[>] Running package
foo


[+] INFO: Packaging pactest
	[>] Packing control.tar
	[>] Packing data.tar
	[>] Compressing
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'pactest' instead of '/usr/src/pacstall/pactest_1.0.0-pacstall1_amd64.deb'
The following NEW packages will be installed:
  pactest
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/806 B of archives.
After this operation, 0 B of additional disk space will be used.
Get:1 /usr/src/pacstall/pactest_1.0.0-pacstall1_amd64.deb pactest amd64 1.0.0-pacstall1 [806 B]
Selecting previously unselected package pactest.
(Reading database ... 16610 files and directories currently installed.)
Preparing to unpack .../pactest_1.0.0-pacstall1_amd64.deb ...
Unpacking pactest (1.0.0-pacstall1) ...
Setting up pactest (1.0.0-pacstall1) ...
[+] INFO: Performing post install operations
	[>] Storing pacscript
[+] INFO: Done installing pactest
[+] INFO: Cleaning up

```

As seen, the scalar variable is propagated into the function, when invoked by Pacstall, but not so the array variables, whether indexed or associative. In contrast, all variables are handled consistently directly by Bash.

The test depends on an arbitrary tar archive, provided locally, which has been called `pactest.tar`.

The environment for the test was a Docker container called `pacstall/ubuntu-noble:master` generated from the following command:

 ````
pacstall-docker-builder --distro ubuntu:noble --version master
 ````

The version of the builder tool is 0.1.3. The host was also based on Ubuntu 24.04, architecture AMD64.

A complete copy of the test case, including the script and a source archive, has been [provided as a Github Gist](https://gist.github.com/brainchild0/fcc0a4badd31e8e6136e12501603788d) (which may be cloned using Git or downloaded as a Zip file).

# Expected behavior

It is expected that all such variables propagate into the function, even when invoked by Pacstall, whether scalar, indexed array, or associative array.

# Notes

The logic for propagation and cleaning of the environment for sandboxed function invocation appears to be deeply convoluted and circuitous. The desired functionality, it seems, certainly could be achieved more transparently, along with corrections for the incomplete propagation of the desired subset of the environment.

One helpful observation is that the shell built-in `declare -v <varname>` prints code that will execute safely in a different shell to set the variable *<varname>* to the exact value as from the original context, including its array attributes and value.

D-Brox:
## Purpose

Properly export the normal and associative arrays to the bwrap functions

## Approach

`env` doesn't list arrays, so `compgen` ftw

## Progress

- [x] Do it
- [x] Test it

## Addendum

Closes #1449

### DIFF
--- a/misc/scripts/bwrap.sh
+++ b/misc/scripts/bwrap.sh
@@ -55,7 +55,7 @@ readonly __OLD_ENV
 $(declare -pf def_colors) && def_colors
 $(for i in {ask,fancy_message,parse_source_entry,calc_git_pkgver}; do declare -pf "${i}"; done)
 source "${input}"
-mapfile -t NEW_ENV < <(comm -13 --nocheck-order <(printf '%s\n' \"\${__OLD_ENV[@]}\") <(compgen -A variable |sort))
+mapfile -t NEW_ENV < <(comm -13 --nocheck-order <(printf '%s\n' \"\${__OLD_ENV[@]}\") <(compgen -A variable | sort))
 declare -p \${NEW_ENV[@]} >> "${bwrapenv}"
 declare -pf >> "${bwrapenv}"
 echo > "${safeenv}"
@@ -102,7 +102,7 @@ mapfile -t OLD_ENV < <(compgen -A variable | sort)
 source ${bwrapenv}
 ${func} 2>&1 "${LOGDIR}/$(printf '%(%Y-%m-%d_%T)T')-$name-$func.log" && FUNCSTATUS="\${PIPESTATUS[0]}" && \
 if [[ \$FUNCSTATUS ]]; then \
-    mapfile -t NEW_ENV < <(comm -13 --nocheck-order <(printf '%s\n' \"\${OLD_ENV[@]}\") <(compgen -A variable |sort)); \
+    mapfile -t NEW_ENV < <(comm -13 --nocheck-order <(printf '%s\n' \"\${OLD_ENV[@]}\") <(compgen -A variable | sort)); \
     declare -p \${NEW_ENV[@]} >> "${bwrapenv}"; \
 fi && ignore_stack=true && exit \$FUNCSTATUS
 EOF
--- a/misc/scripts/bwrap.sh
+++ b/misc/scripts/bwrap.sh
@@ -50,13 +50,12 @@ function safe_source() {
 
     sudo tee "$tmpfile" > /dev/null << EOF
 #!/bin/bash -a
-mapfile -t __OLD_ENV < <(compgen -A variable  -P "--unset ")
+mapfile -t __OLD_ENV < <(compgen -A variable | sort)
 readonly __OLD_ENV
 $(declare -pf def_colors) && def_colors
 $(for i in {ask,fancy_message,parse_source_entry,calc_git_pkgver}; do declare -pf "${i}"; done)
 source "${input}"
-mapfile -t NEW_ENV < <(/bin/env -0 \${__OLD_ENV[@]} | \
-    sed -ze 's/BASH_FUNC_\(.*\)%%=\(.*\)\$/\n/g;s/^\(.[[:alnum:]_]*\)=\(.*\)\$/\1/g'|tr '\0' '\n')
+mapfile -t NEW_ENV < <(comm -13 <(printf '%s\n' \"\${__OLD_ENV[@]}\") <(compgen -A variable |sort))
 declare -p \${NEW_ENV[@]} >> "${bwrapenv}"
 declare -pf >> "${bwrapenv}"
 echo > "${safeenv}"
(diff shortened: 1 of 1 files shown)


oklopfer:
@D-Brox can you explain what unset was doing before and also what the sed was actually doing

D-Brox:
It was a prefix to the `compgen` list responses. It was passed to the `env` as the unset flag to filter out the old env variables

The sed just extracted the variable names.

oklopfer:
Still need to test but ok 😃👍

misc/scripts/bwrap.sh
```diff
@@ -50,13 +50,12 @@ function safe_source() {
...
 #!/bin/bash -a
-mapfile -t __OLD_ENV < <(compgen -A variable  -P "--unset ")
+mapfile -t __OLD_ENV < <(compgen -A variable | sort)
 readonly __OLD_ENV
 $(declare -pf def_colors) && def_colors
 $(for i in {ask,fancy_message,parse_source_entry,calc_git_pkgver}; do declare -pf "${i}"; done)
 source "${input}"
-mapfile -t NEW_ENV < <(/bin/env -0 \${__OLD_ENV[@]} | \
-    sed -ze 's/BASH_FUNC_\(.*\)%%=\(.*\)\$/\n/g;s/^\(.[[:alnum:]_]*\)=\(.*\)\$/\1/g'|tr '\0' '\n')
+mapfile -t NEW_ENV < <(comm -13 <(printf '%s\n' \"\${__OLD_ENV[@]}\") <(compgen -A variable |sort))
```
oklopfer: https://github.com/pacstall/pacstall/commit/910748f97258078a86db954b466810c13a374e5b

Need flag

Also space should be between pipe and sort

oklopfer (changes requested):

misc/scripts/bwrap.sh
brainchild0: I have experience with failures in shell scripts due to the inputs provided for `comm` both having have been sorted, but whose sorting has been following comparison rules for different locales. The problem is exacerbated because the locale applied for sorting may not be determined by the same factors across all sort operations, which may be performed in various processes other than the core utility `sort`.

 It may be most robust, even if only as a precaution, to sort all lists explicitly, (that is, inserting `sort` into the pipe), while forcing the sort locale to `C`, through the appropriate and consistent use of variables `LC_ALL` and `LC_LOCALE` in the environments for all invocations of `sort`. 

It may seem cumbersome. However, please remember the convenience that an alias or function takes precedence over a executable file in the search path, as commands are resolved by Bash.

oklopfer: You should just set your locale to C if it is causing you problems.

oklopfer (approved):
seems to work

misc/scripts/bwrap.sh
brainchild0: You are completely missing the point.

We write software that functions as we want under the broadest range of conditions, not that breaks readily unless the conditions conform to some ideal.

1. The application should function correctly regardless of the inherited environment.
2. Not all sorting procedures among various utilities resolve the locale by the same means.
3. The environment that propagates down the process tree cannot be predicted reliably by any means that are transparent or straightforward.  
4. Unless each component is independently robust, subtle modifications to the software that may be introduced later, and that seem harmless, may expose latent vulnerabilities, that cause breakage under special cases that extremely difficult to trace.

PR #1452, opened 2026-03-27
feat(fetch-sources.sh): Improve display of source names 

brainchild0:
Currently, while a script is processed, messages about actions on sources, such as "Downloading" and "Extracting", are not consistent across different cases, or necessarily strongly helpful to identify each of the sources.

If a script provides no named destination (i.e. `dest::` prefix) for the source field, then the file name or repository name, stripped of the full address, is displayed in highlighted text.

Such behavior seems generally strongly desirable.

However, if a script names a destination, then the name is shown instead of the actual source information. The destination name is often opaque for the end user, chosen as an implementation detail of the script, best to facilitate its maintenance. It may be exceedingly terse, to prevent source fields from appearing overly cluttered or redundant in the script specifications.

From the standpoint of reporting major operations processing the source, the destination name is irrelevant and confusing.

The commit makes the behavior correct and consistent in either case.

---

Note, please, that I have not found a testing framework in the codebase, nor a description of testing processes in basic documentation. The revisions have been casually tested for correctness, but I rely on further guidance to understand the expected methods for testing, that the revisions be considered seriously for integration into the project.

---

*I confirm that I have read the [contributing guidelines](https://github.com/pacstall/pacstall/blob/develop/CONTRIBUTING.md), and this pull request is abiding by all the clauses stated in the guideline.*

### DIFF
--- a/misc/scripts/fetch-sources.sh
+++ b/misc/scripts/fetch-sources.sh
@@ -66,6 +66,7 @@ function parse_source_entry() {
     if [[ ${dest} == *"?"* ]]; then
         dest="${dest%%\?*}"
     fi
+    src_dsply="${source_url##*/}"
 }
 
 function calc_git_pkgver() {
@@ -221,18 +222,18 @@ function git_down() {
     if [[ -n ${git_branch} || -n ${git_tag} ]]; then
         if [[ -n ${git_branch} ]]; then
             revision="${git_branch}"
-            fancy_message info $"Cloning %b from branch %b" "${BPurple}${dest}${NC}" "${CYAN}${git_branch}${NC}"
+            fancy_message info $"Cloning %b from branch %b" "${BPurple}${src_dsply}${NC}" "${CYAN}${git_branch}${NC}"
         elif [[ -n ${git_tag} ]]; then
             revision="${git_tag}"
-            fancy_message info $"Cloning %b from tag %b" "${BPurple}${dest}${NC}" "${CYAN}${git_tag}${NC}"
+            fancy_message info $"Cloning %b from tag %b" "${BPurple}${src_dsply}${NC}" "${CYAN}${git_tag}${NC}"
         fi
         gitopts="-b ${revision}"
     elif [[ -n ${git_commit} ]]; then
         gitopts=("--no-checkout" "--filter=blob:none")
-        fancy_message info $"Cloning %b with no blobs" "${BPurple}${dest}${NC}"
+        fancy_message info $"Cloning %b with no blobs" "${BPurple}${src_dsply}${NC}"
     else
         unset gitopts
-        fancy_message info $"Cloning %b from %b" "${BPurple}${dest}${NC}" "${CYAN}HEAD${NC}"
+        fancy_message info $"Cloning %b from %b" "${BPurple}${src_dsply}${NC}" "${CYAN}HEAD${NC}"
     fi
     # git clone quietly, with no history, and if submodules are there, download with 10 jobs
     # shellcheck disable=SC2086,SC2031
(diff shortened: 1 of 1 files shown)


oklopfer:
No, this is our desired intentions.

brainchild0:
Your intentions needlessly confuse the end user.

Display output should not be dependent on implementation details of the script.

The proposed commit enhances usability, without any doubt.

I find it confounding that you would not reconsider.

oklopfer:
It tells the end user how it is seen by the system and where it is going, because that is actually what matters. Source file names are equally as arbitrary as script defined ones.

brainchild0:
*I know you prefer brevity, but I hope you are willing to consider the nuanced differences in our perspectives.*

*To help prevent a sense of excessive complexity, I have added emphasis to the most essential passages.*

---

When considering information reported to the user, several related but distinct considerations are relevant.

Certainly, which information to report must be considered. Further, it must be considered when, where, and how for it to be reported. 

A final consideration that must not be overlooked, obviously, is whether any choice would result in reporting that is inaccurate.

Assessing accuracy includes assessing, in particular, differences in the interpretations and inferences reached by developer versus user, as well as one against another user. It is not suitable for the developer's decisions to be based on an assumption that all users think and act as the developer, or that all users invoke the same operation while having the same information and interests. 

---

*Now, to begin discussing the current behavior, the fact is that by any standard, some of the reporting is inaccurate.*

The message literally includes the word "downloading" followed by a name that is not the name of an item being downloaded. If the destination is to be reported, then it must be reported as the destination, not, as is currently the case, reported as the source.

Further, if the destination is reported, then the archive extensions from the source should not be included, as currently is the case, because the destination, in actual fact is stripped of such information. Clearly, the actual destination, as such, also should be the destination reported, if reporting the destination is to have any reliable utility. Therefore, this second feature of the reporting may be included among those that are plainly inaccurate.

*In short, the two cases we observe are, one, reporting a name as being downloaded that is not being downloaded, and, the other, confusingly reporting as the destination a name that is not even always the destination.*

*Surely, something needs to change.*

---

As it happens, the later extraction phase is the one in which the destination is most strongly relevant. If a destination is to be reported, then it is more sensible that it be reported for extraction than for downloading. 

Whereas the reporting for downloading clearly should never omit the actual source being downloaded, the reporting for extraction entails a choice of three options, which are, to report only the source, only the destination, or both the source and the destination. Since the source is to be reported for downloading, the corresponding source also should be reported as it is being extracted, for clarity and continuity, even if the the corresponding destination also is reported. Reporting only the destination is inadequate.

Therefore, to accommodate all of the various concerns, including yours, it may be most sensible to include both the source and destination in the reporting of the extraction.

Now, the final and more abstract point of controversy is whether the destination is ever to be reported. I argue that reporting the destination is unnecessary, distracting, and confusing.

I am attempting to complete a script, and for reasons mentioned elsewhere, I am essentially forced to use an explicit destination. The source addresses are very long, due to excessive verbosity and redundancy in their naming. To keep the sources section of the script readable, I use a single letter for the destination. The choice is my own implementation detail that simplifies script maintenance and has no consequences for the end user. The end user is concerned, at most, with the sources being downloaded for processing. and then the processing applied to the contents extracted from the sources. Such are the essential features of building each packages as different from building another package.

*The destination is an arbitrary intermediary that offers no value or difference to the end user. While the destination is arbitrary, the source is not arbitrary, because the script cannot designate an arbitrary source while still producing the same correct behavior. The source must be consistent with its location in the world, outside the script, whereas the destination must only be consistent internally to the script. Despite your objection, the source is not arbitrary, the same as the destination being arbitrary.*

If the user is a script maintainer, then reporting the destination may be useful, as well as certainly is useful to report the source. I find no other case in which the destination ever may be plausibly interesting or consistently meaningful to the user.  Due to such differences, I might repeat my earlier suggestion that separate levels of verbosity be supported, one targeting using scripts by the end user, versus the other targeting testing scripts by the maintainer. As already explained elsewhere, such relatively minor differences in verbosity are quite distinct from the ones that are included by the current debug feature, whose intended use is quite different. Regardless of whether such a feature ever is introduced, the above argument for certain changes being necessary remains sound, and such is essential to consider.

oklopfer:
You should think of $dest as an alias for a file being downloaded, not an exact naming. Imagine a package that downloads two files, both named `archive.zip`. $dest is used to prevent conflict in this case. If it reports "Downloading `archive.zip`" twice, that is confusing. It should use the alias in the reporting.

Please consider this the end of this conversation, there will be no budging; it is very intentionally designed. I will not be responding to further replies in this thread.

PR #1454, opened 2026-03-28
feat(build.sh): Optimize build times when not keeping package file

brainchild0:
The details for building a package depend on whether the package file will be kept after installation, such as to be suitable for external distribution, or instead, deleted immediately following completion of successful system installation. 

The former is optimized for storage space, the latter for build speed. 

The faster but larger case currently relies on Gzip compression, but such is unnecessary. Modern and recent revisions of the Debian package system naturally process tar archives, within the base package archive, that are uncompressed. 

Whereas the Gzip case, with compression options optimized for speed, is noticeably more time efficient than the alternative case, Zstandard compression optimized for space, the uncompressed case remains yet vastly still superior, and such differences are significant to the user when processing larger packages. 

The revision vastly reduces processing times for larger packages.

### DIFF
--- a/misc/scripts/build.sh
+++ b/misc/scripts/build.sh
@@ -303,45 +303,65 @@ function createdeb() {
     local debname="${1}_${2}_${3}"
     if ((PACSTALL_INSTALL == 0)); then
         # We are not going to immediately install, meaning the user might want to share their deb with someone else, so create the highest compression.
-        local flags=("-19" "-T0" "-q")
-        local compression="zst"
-        local command="zstd"
+        local cmd_comp="zstd"
+        local flags_comp=("-19" "-T0" "-q" "--rm")
+        local ext_comp="zst"
     else
         # Immediate install (gzip), so we want fast build times over everything else
-        local flags=("-1n")
-        local compression="gz"
-        local command="gzip"
+        local cmd_comp="gzip"
+        local flags_comp=("-1n")
+        local ext_comp="gz"
     fi
+
     cd "$STAGEDIR/$pacname" || { ignore_stack=true; return 1; }
     # https://tldp.org/HOWTO/html_single/Debian-Binary-Package-Building-HOWTO/#AEN66
+
+    local debfile="../$debname.deb"
+
     echo "2.0" | sudo tee debian-binary > /dev/null
-    sudo tar -cf "$PWD/control.tar" -T /dev/null
-    local CONTROL_LOCATION="$PWD/control.tar"
+
+    shopt -s nullglob
+
+    local cname="control.tar.$ext_comp"
+    local cpath="$(pwd)/$cname"
+
     # avoid having to cd back
     pushd DEBIAN > /dev/null || { ignore_stack=true; return 1; }
-    for i in *; do
-        if [[ -f $i ]]; then
-            local files_for_control+=("$i")
+    local filename
+    for filename in *; do
+        if [[ -f $filename ]]; then
+            local citems+=("$filename")
         fi
     done
-    fancy_message sub $"Packing control.tar"
-    sudo tar -rf "$CONTROL_LOCATION" "${files_for_control[@]}"
-    popd > /dev/null || { ignore_stack=true; return 1; }
-    sudo tar -cf "$PWD/data.tar" -T /dev/null
-    local DATA_LOCATION="$PWD/data.tar"
-    # collect every top level file/dir except for deb stuff
-    for i in *; do
-        [[ $i =~ ^(DEBIAN|control.tar|data.tar|debian-binary)$ ]] && continue
-        local files_for_data+=("$i")
-    done
-    fancy_message sub $"Packing data.tar"
-    sudo tar -rf "$DATA_LOCATION" "${files_for_data[@]}"
 
-    fancy_message sub $"Compressing"
-    sudo "$command" "${flags[@]}" "$DATA_LOCATION" "$CONTROL_LOCATION"
-    sudo ar -rU "$debname.deb" debian-binary control.tar."$compression" data.tar."$compression" > /dev/null 2>&1
-    sudo mv "$debname.deb" ..
-    sudo rm -f debian-binary control.tar."$compression" data.tar."$compression"
+    fancy_message sub $"Packing and compressing $cname"
+
+    sudo tar -c "${citems[@]}" \
+        | "$cmd_comp" "${flags_comp[@]}" \
+        | sudo tee "$cpath" > /dev/null
+
+    popd > /dev/null || { ignore_stack=true; return 1; }
+
+    sudo ar qc "$debfile" debian-binary "$cname"
+
+    # Exclude control metadata from package contents
+    sudo rm -r DEBIAN debian-binary "$cname"
+
+    local dname="data.tar.$ext_comp"
+    local dpath="$(pwd)/$dname"
+
+    fancy_message sub $"Packing and compressing $dname"
+
+    sudo tar -c * .* \
+        | "$cmd_comp" "${flags_comp[@]}" \
+        | sudo tee "$dpath" > /dev/null
+
+    sudo ar qc "$debfile" "$dname"
+
+    sudo rm "$dname" > /dev/null 2>&1
+
+    shopt -u nullglob
+
 }
 
 function is_builddep_arch() {
--- a/misc/scripts/build.sh
+++ b/misc/scripts/build.sh
@@ -306,11 +306,13 @@ function createdeb() {
         local cmd_comp="zstd"
         local flags_comp=("-19" "-T0" "-q" "--rm")
         local ext_comp="zst"
+        local pack_msg_fmt="Packing and compressing %s"
     else
-        # Immediate install (gzip), so we want fast build times over everything else
-        local cmd_comp="gzip"
-        local flags_comp=("-1n")
-        local ext_comp="gz"
+        # Immediate install, so we want fast build times over everything else
+        local cmd_comp=cat
+        local flags_comp=()
+        local ext_comp=
+        local pack_msg_fmt="Packing %s"
     fi
 
     cd "$STAGEDIR/$pacname" || { ignore_stack=true; return 1; }
@@ -322,7 +324,7 @@ function createdeb() {
 
     shopt -s nullglob
 
-    local cname="control.tar.$ext_comp"
+    local cname="control.tar${ext_comp:+.${ext_comp}}"
     local cpath="$(pwd)/$cname"
 
     # avoid having to cd back
@@ -334,7 +336,7 @@ function createdeb() {
         fi
     done
 
-    fancy_message sub $"Packing and compressing $cname"
+    fancy_message sub "$(printf "$pack_msg_fmt" "$cname")"
 
     sudo tar -c "${citems[@]}" \
         | "$cmd_comp" "${flags_comp[@]}" \
@@ -347,10 +349,10 @@ function createdeb() {
     # Exclude control metadata from package contents
     sudo rm -r DEBIAN debian-binary "$cname"
 
-    local dname="data.tar.$ext_comp"
+    local dname="data.tar${ext_comp:+.${ext_comp}}"
     local dpath="$(pwd)/$dname"
 
-    fancy_message sub $"Packing and compressing $dname"
+    fancy_message sub "$(printf "$pack_msg_fmt" "$dname")"
 
     sudo tar -c * .* \
         | "$cmd_comp" "${flags_comp[@]}" \


misc/scripts/build.sh
```diff
@@ -303,45 +303,67 @@ function createdeb() {
...
+    local filename
+    for filename in *; do
+        if [[ -f $filename ]]; then
+            local citems+=("$filename")
         fi
     done
-    fancy_message sub $"Packing control.tar"
-    sudo tar -rf "$CONTROL_LOCATION" "${files_for_control[@]}"
+
+    fancy_message sub "$(printf "$pack_msg_fmt" "$cname")"
```
Elsie19, line 339:
```suggestion
    fancy_message sub "$pack_msg_fmt" "$cname"
```

This also has to be a gettext string (`$"like this"`)

Elsie19 (changes requested):

misc/scripts/build.sh
```diff
@@ -303,45 +303,67 @@ function createdeb() {
...
+
+    sudo ar qc "$debfile" debian-binary "$cname"
+
+    # Exclude control metadata from package contents
+    sudo rm -r DEBIAN debian-binary "$cname"
+
+    local dname="data.tar${ext_comp:+.${ext_comp}}"
+    local dpath="$(pwd)/$dname"
+
+    fancy_message sub "$(printf "$pack_msg_fmt" "$dname")"
```
Elsie19, line 355: Same as above.

Elsie19:
Has this been tested?

D-Brox:
The idea is sound, but the code doesn't match, both quality-wise and logic-wise. 

The team will re-implement this PR soon, taking into account both the different possible use cases and our code standard.

brainchild0:
I could have provided the new functionality with only trivial changes.

Instead, I also dramatically improved the original code, which was low quality. It was difficult to read, poorly structured, and, due to very noticeable weakness in the style and logic, appeared prone to failure in a variety of fringe cases.

I did you a favor by rewriting the whole function rather than doing the minimum work to achieve the new functionality. The branch is divided into two commits. The first provides no new functionality, but still offers you an improvement to your project. I suggest considering it as a basis of further improvement.

The new code is easy to follow as correct and reliable. It is possible that further testing would reveal a regression, but if so, it would be even more straightforward to correct, following the improved quality.

If you have differences over style that you wish to address, then I am completely willing to discuss adjustments. However, I not only feel troubled by your unwillingness to be receptive to my efforts, but more importantly, I strongly question the wisdom of rejecting the more general contributions.

D-Brox:
Your implementation attempt does not take into account the K and B flags, which is why the team has decided that we will be reimplementing this PR with some extra use cases in mind.

Again, the idea was sound, but the execution lacks.

misc/scripts/build.sh
brainchild0: I have not used `gettext`.

A format string abstracts the sequencing of static and parameterized elements that generate the output. They allow modifications to the presentation without affecting the logic or structure.

Have you considered?:

```
fmt_str="$(gettext "...")"
```

oklopfer:
FWIW looking back L304 `if ((PACSTALL_INSTALL == 0)); then` should keep `-B` working properly, but `-K` still needs proper implementation, and there should not be so many variables rewritten here

I don't disagree that this command could use a rework, it is messy, but that can be done without as many major changes.

brainchild0:
1. My refactoring reworks how the package is built. Whether the package is kept, on the filesystem, is resolved elsewhere.
2. If I am mistaken, or if you have found any regression, then please bring it to my attention, and give me an opportunity to respond. The best correction very likely would be a further improvement or appropriate correction, not disposal, of the structure I have conceived.
3. "Major changes' should not be feared. In many contexts, of course, changes should be limited, to protect against the possibility of breakage. However, the "major changes" I have written may be seen as generally clear and sound. Again, if a regression has been identified, then it can be addressed within the context of the new structure.  My tests reveal that the logic is, in the larger balance, correct. A project cannot advance if "major changes" are considered as totally unacceptable in all cases, and original choices are generally treated as sanctimonious. 
4. Rewriting variables is often an effective improvement. Most software is written originally with an emphasis on achieving functionality by any means convenient. Once the functionality is secure, rethinking the overall structure for readability and maintainability is necessary for advancement. Again, the earlier form should not be defended as sanctimonious.
5. The quality of this particular section of code was quite poor, to be frank. It was not robust against fringe conditions, nor clear to understand, nor planned as predictably reliable in isolation from other components. Any refactoring into a generally cogent form should be recognized as an improvement, even if some objections are raised, or further improvements are desired. I am not criticizing those who labored to achieve a functional result, but I am challenging those who seem to be rebuking further progress.

oklopfer:
See the PR linked above, which is a necessary precursor before any next steps. Please also take a look at the description I gave in that PR, which clearly lays out exactly what kind of changes I made, in short bullet points. You should be doing the same, otherwise it makes it difficult to know what your code is supposed to be doing.

brainchild0:
My objection is not that you maintain reservations, or that you favor further development, against my proposed contributions.

My objection is that even while I provide contributions that you acknowledge as meaningful, or that are provably correct by isolated test cases, even against claims to the contrary, I have been treated as someone who at best can only aspire to approximate your sublime wisdom.

My motive is not to prove you wrong, or to establish myself as superior, only for it to be acknowledged that no one knows everything.

You are not repeating the same explanations. You are repeating your own entrenched dogma while willfully rejecting any suggestion that your dogma ought to be reconsidered.

You are not protecting a fragile codebase. You are alienating those who have an interest and capability in contributing to its gradual further stability.

I make no claim of expertise in all the same subjects as have you, but I reject that you know better than I in every possible case.


At present, I have limited interest in the PR you mention, because the overall handling of the PR I have opened has been oriented to exclude me from the process initiated by my own insight and effort.  

My PR is a perfectly suitable location to discuss the particular substance of my PR.

In fact, proposing revisions to a PR, and adding written comments, are both features specifically supported by the platform.

Why didn't you respond to my PR by clarifying the specific aspects of your approval or concern, such as to open an a opportunity for discussion, and asking me to evaluate your own proposed further revisions?





Against an opportunity to collaborate respectfully toward common interests, you instead dismissed my work as "crap".

misc/scripts/build.sh
Elsie19: > I have not used `gettext`.
> 
> A format string abstracts the sequencing of static and parameterized elements that generate the output. They allow modifications to the presentation without affecting the logic or structure.
> 
> Have you considered?:
> 
> ```
> fmt_str="$(gettext "...")"
> ```
> 
> 

No, you will use `$""` or you will not print a string at all. That is non negotiable. `fancy_message` supports `printf` style format strings, such as:

```bash
fancy_message sub $"hello %s!" "world"
```

D-Brox:
Look, the old code, as in the existing code not the one from the PR, was tech debt crap (it exists specifically because zstd wasn't supported by apt on Debian until last year).

We are trying to avoid the use of indiscriminate globbing in the newer code like the `tar -c * ./*` in your PR.

Renaming the variables from something descriptive `CONTROL_LOCATION` to `cpath` makes it less readable.

We are not against changes, but the more complex ones must be thoroughly tested. You have already helped us a lot with the issues you opened, part of the reason I argued for your unblocking in the first place. Eyes on the code is always good. 

Pacstall is full of tech debt, I won't deny it, and that's why we want to rewrite it in another language, bash is not ideal for a package manager. In the last 2 years we tried modernizing the bash codebase, but things always slip amongst the lines. We've been working on pacstall for upwards of 6 years, so we know more of the weird/unusual quirks of the codebase.

I have nothing against you or your insights, but again, please, use the issue and PR templates, they make the initial communication easier.

D-Brox:
Communication is a two way street, so try to be plain and direct, being overly verbose hinders communication, reduces clarity, and makes it harder for the reader.

While your verbosity may stem from a desire to be thorough, excessive verbosity is often perceived as pretentious, arrogant, or an attempt to mask a lack of clear thinking.

misc/scripts/build.sh
brainchild0: I believe `gettext` and `$""` are equivalent. Regardless, the same concept may be applied.


```
f='hello %s!'
printf $"$f\n" world
```

Elsie19: Yes, `gettext` and `$""` are equivalent, but that is irrelevant. We use `$""`, not `gettext`. Secondly, your example is actually not correct. That produces this POT file:

```pot
#: test.sh:2
msgid "$f"
msgstr ""
```

Which is not useful at all, because it only contains a variable and no text, and even besides all that, you should not be interpolating variables *inside* `printf`. The correct way, as I have said before, is to do:

```bash
fancy_message sub $"hello %s!" "world"
```

Which produces this file:

```pot
#: test.sh:1
msgid "hello %s!"
msgstr ""
```

brainchild0: > that is irrelevant

It may seem irrelevant if you are quibbling over details rather than analyzing the concept as it could be applied in either case.

> The correct way, as I have said before

It would work only if `fancy_message` processed printf arguments, which though is not the case, could be implemented fairly trivially.

Elsie19: > It may seem irrelevant if you are quibbling over details rather than analyzing the concept as it could be applied in either case.

What are you even saying? WE USE GETTEXT STRINGS, NOT THE EXTERNAL GETTEXT PROGRAM!!!!!! ***WE. DON'T. DO. IT. YOUR. WAY. PERIOD.***

> It would work only if `fancy_message` processed printf arguments, which though is not the case, could be implemented fairly trivially.

Dude. https://github.com/pacstall/pacstall/blob/2f5d0cb49a548c13a263770d428c78a6917619fe/pacstall#L179-L194

Have you even *read* our codebase?

brainchild0:
```
$ xgettext -L shell - -o - <<"EOF"
fmt_str="$(gettext 'hello %s\n')" 
EOF
# SOME DESCRIPTIVE TITLE.
# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
# This file is distributed under the same license as the PACKAGE package.
# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
#
#, fuzzy
msgid ""
msgstr ""
"Project-Id-Version: PACKAGE VERSION\n"
"Report-Msgid-Bugs-To: \n"
"POT-Creation-Date: 2026-03-29 14:55-0400\n"
"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
"Language-Team: LANGUAGE <LL@li.org>\n"
"Language: \n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=CHARSET\n"
"Content-Transfer-Encoding: 8bit\n"

#: standard input:1
msgid "hello %s\\n"
msgstr ""
```

brainchild0: > Have you even _read_ our codebase?

My mistake.

brainchild0: > _**WE. DON'T. DO. IT. YOUR. WAY. PERIOD.**_

My point was that the concept I discussing  is the same regardless of the choice of `gettext` versus `$""`.

Which one I used merely in my example was not relevant, because the details I was discussing are the same in either case.

brainchild0:
Verbosity can cause problems, as you say.

Being overly terse and dogmatic also tends to be problematic.

The comments I receive in this project usually provide no information about the individual's beliefs or reasoning on a subject I try to discuss. My verbosity follows from the requirement that I address all the possible reasons for someone reaching a different conclusion, due to the actual reasons never being revealed. 

What does seem to be revealed, quite frankly, is that the depth of consideration being applied to my specific suggestions is generally quite limited, and the knowledge being applied is not always complete or inaccurate. Yet, the responses still seem to be framed around an inflexible conviction.

It would be best for others to lay out their own argument and understanding, even if only briefly, trying to reach the crux of a disagreement through dialog that is generative, in contrast to the the tendency I have perceived of others acting as assured and dismissive.

For example, it was not constructive that two distinct individuals insisted that bubblewrap blocks standard I/O, when the documentation makes no such claim, and proving the contrary was trivial.

A more constructive answer at least might have resembled the following:

> That idea is interesting, and might be viable if we could establish pipes between the host and isolated child process, but I'm afraid the isolation created through bubblewrap precludes using pipes. I don't think the idea can work, unfortunately, due to the current form of isolation.

Even more helpful might have been one resembling either of the following, as whichever would be applicable:

- > I have never tried using pipes with processes in bubblewrap, but I'm not convinced it would work.

- > I have tried using pipes with bubblewrap, and it never worked. I don't know right now whether using pipes is possible without harmful effects.

brainchild0:
> Look, the old code, as in the existing code not the one from the PR, was tech debt crap (it exists specifically because zstd wasn't supported by apt on Debian until last year).


Yet you made no acknowledgement and showed no receptivity to my efforts to improve the quality. Instead, it seems you dismissed my contribution entirely, and it seems also that you did so largely based on specific concerns that you never mentioned, that you never allowed me to address, and that I believe are most likely specific details as distinct from the broader effort.

> 
> We are trying to avoid the use of indiscriminate globbing in the newer code like the `tar -c * ./*` in your PR.



The original code also used globs just as indiscriminately. I at least did so more robustly. You respond as though I regressed the code quality by introducing further use of globbing. I agree that globs should not be necessary for this operation, but adequately correcting all of the related problems was beyond the scope of the particular contribution. 


Using `tar -C .` is definitely better than `tar . .*`, but lacking this choice should be noticed as an omissions for further revision, not used as a reason to discard all of the changes.

My contribution was intended as a an incremental improvement, within a particular scope, that was unlikely to cause a regression. Writing the function optimally would depend on broader changes, and also carefully assessing whether specific parts of the existing logic could be safely simplified without changing the results in any cases.

Incremental improvements, in succession, is a highly successful means of achieving broader changes, not an effect that works to the contrary.

> 
> Renaming the variables from something descriptive `CONTROL_LOCATION` to `cpath` makes it less readable.


I respect your preference, but the matter is no so unequivocal as you suggest. Local variables have different requirements than variables in a higher scope. Local variables are assigned nearby and invoked frequently. Descriptive names are not necessary. Long names often force lines to overwrap, and statements to span multiple lines. All caps is hard to read and even harder to type.

*Regardless, if you object to the new variable names, then please state the objection. If you think it necessary, then request that they be changed, with or without submitting your own changes.*


> 
> We are not against changes, but the more complex ones must be thoroughly tested. 

Testing changes is the next step after writing changes. Testing changes is essential, but writing changes necessarily occurs first.




> 
> I have nothing against you or your insights, but again, please, use the issue and PR templates, they make the initial communication easier.

Templates are good for basic bugs. I have often followed them, with modifications that I feel are useful in specific cases. Proposing design changes that are more complex and abstract may benefit from a different form, a structured argument that analyzes the deeper nuances.

misc/scripts/build.sh
brainchild0: I misunderstood the conversation. 

 Based on the revisions alone, it seemed to me that the formatting was being removed. I also incorrectly inferred that you were advocating against using both a formatting string and translation mechanism in combination.



For the record, a simple note that printf is used by fancy_message would have been a tremendous help.


Regardless, I apologize for misunderstanding.

One useful result of the conversation, at least, is that there are some cases, which often include formatting strings, in which shell interpolation is not desirable. In such cases, explicitly calling gettext an a single-quoted string is a useful alternative to $"".

Elsie19: > I misunderstood the conversation.

Understood 👍🏻

> One useful result of the conversation, at least, is that there are some cases, which often include formatting strings, in which shell interpolation is not desirable. In such cases, explicitly calling gettext an a single-quoted string is a useful alternative to $"".

Can you give an example? We have never encountered something where we would ever need to explicitly call `gettext`.

oklopfer:
You seem to be largely ignoring what we are saying to you, so I am going to point you to look at #1458 again, which is performing cleanup in order for any of us to enact this compression change. For example, the globbing is removed within that PR, along with other unnecessary gunk; that gunk is what I am referring to as crap.

Nothing from that PR negates what you are trying to do here in a broader scope (in fact I specifically mention at the top that it is going to be beneficial for achieving the goals here), but this will need some redoing once that PR is merged. Nonetheless, it is better for all of us to separate the cleanup and refactor into two steps. We are plenty guilty of combining those two processes (cleaning up/trimming down and refactoring) in the past, and it has come back to bite us in the ass when trying to retrace why code was introduced when new problems arise.

Continuing to name call and cry foul play is not going to help your case and does not motivate us to work with you in any way. You are trying to tell us on our own project how we should be writing and assessing code. If you want to contribute, it is an absolute and non-negotiable requirement that you abide by our standards, not the other way around. For everyone's sake, please take a step back and stop arguing. You are convincing no one.

misc/scripts/build.sh
brainchild0:
```
$ bash<<"EOF"
f1=$"%s paid $10 for lunch.\n"
f2=$(gettext '%s paid $10 for lunch.\n')

printf "$f1" "elsie"
printf "$f2" "elsie"

EOF
elsie paid 0 for lunch.
elsie paid $20 for lunch.
```

---

Also, newer versions of `printf` support the `$` as a formatting element, used in the recently introduced support for repositioning parameters from the order that they follow the formatting string.

For such an invocation as follows, the output would be "*Hello world!*".

```
printf '%2$s %1$s!\n' "world" "Hello"
```

---

The workaround for the issue with double quotes is the backslash escape of `$`, but it degrades readability and maintainability, especially in the more complicated case or repositioning.

As least since Bash applies no special processing to other common backslash-escape sequences, the particular issue is limited to fringe cases.

In contrast, Perl has both a version of `printf`, and a much stronger distinction between double versus single quotes.

PR #1458, opened 2026-03-29
fix(build.sh): reduce complexity for further refactoring

oklopfer:
## Purpose

reduce the crap to make the goals of #1454 easier to achieve

## Approach

- put locals on single lines
- drop the need for push and pop
- drop the cycling to collect files
  - use protective find rather than globbing

## Progress

- [x] do it
- [x] test it

## Checklist

### DIFF
--- a/misc/scripts/build.sh
+++ b/misc/scripts/build.sh
@@ -300,42 +300,25 @@ function clean_logdir() {
 
 function createdeb() {
     { ignore_stack=false; set -o pipefail; trap stacktrace ERR RETURN; }
-    local debname="${1}_${2}_${3}"
+    local debname="${1}_${2}_${3}" CONTROL_LOCATION="$STAGEDIR/${1}/control.tar" DATA_LOCATION="$STAGEDIR/${1}/data.tar"
     if ((PACSTALL_INSTALL == 0)); then
         # We are not going to immediately install, meaning the user might want to share their deb with someone else, so create the highest compression.
-        local flags=("-19" "-T0" "-q")
-        local compression="zst"
-        local command="zstd"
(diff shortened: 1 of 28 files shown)


oklopfer:
I'll reduce `${STAGEDIR}/${1}` to a local variable too, after some rest

misc/scripts/build.sh
```diff
@@ -300,42 +300,26 @@ function clean_logdir() {
...
+
+    fancy_message sub $"Packing %s" "data.tar"
     # collect every top level file/dir except for deb stuff
-    for i in *; do
-        [[ $i =~ ^(DEBIAN|control.tar|data.tar|debian-binary)$ ]] && continue
-        local files_for_data+=("$i")
-    done
-    fancy_message sub $"Packing data.tar"
-    sudo tar -rf "$DATA_LOCATION" "${files_for_data[@]}"
+    find . -maxdepth 1 -mindepth 1 ! -name DEBIAN ! -name control.tar ! -name data.tar ! -name debian-binary -print0 \
```
Elsie19, line 321: Do these need to be backslash protected?

oklopfer: Why would they need to be?

Elsie19: Because of `!`.

```man
RESERVED WORDS
       Reserved  words  are  words that have a special meaning to the shell.  The following
       words are recognized as reserved when unquoted and either the first word of  a  com‐
       mand  (see SHELL GRAMMAR below), the third word of a case or select command (only in
       is valid), or the third word of a for command (only in and do are valid):

       ! case  coproc  do done elif else esac fi for function if in select then until while
       { } time [[ ]]
```

Elsie19 (approved):

PR #1461, opened 2026-03-29
feat(main): add `PACSTALL_XTRACE{FD,LOG}` for debugging

### LINKED ISSUE #1457: bugs: stacktrace and debug not working correctly
### Describe the bug

BASH_XTRACEFD not work with elevate
Also stacktrace isn't catching in the places where it used to, no bueno

### To reproduce

debug: `BASH_XTRACEFD=3 pacstall -x -I capsfix-git 3>/tmp/pac-debug.log`
stacktrace: `"gzip"` > `"gzip iiiiiiii"` in build.sh

### Expected behavior

debug: should be printing all of it to the log with xtrace, not the console
stacktrace: should be catching whenever failure (oddly, it catches in other places, like making `pacfile` an invalid file in `package-base.sh`); it used to catch on that intentional compression type error, but no longer; why inconsistent?

### Screenshots

later

### OS version

Rhino 2025.4

### Pacstall version

6.3.11

D-Brox:
## Purpose

Fix the redirect with xtrace in elevated instance

## Approach

Add `PACSTALL_XTRACELOG` and `PACSTALL_XTRACEFD` variables used in conjunction with the debug flag

## Progress

- [x] Do it
- [x] Test it

## Addendum
Partially fixes #1457

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -46,6 +46,12 @@ export PACSTALL_INSTALL=1
 
 PACSTALL_DEBUG=0
 if [[ "${1}" == "-x" || "${1}" == "--debug" ]]; then
+  if [[ -v PACSTALL_LOG ]]; then
+    # Separate elevated log file because of permissions issues
+    [[ $PACSTALL_ELEVATED ]] && PACSTALL_LOG="${PACSTALL_LOG}.elevated"
+    exec 3>>"${PACSTALL_LOG}"
+    export BASH_XTRACEFD=3
+  fi
   set -x
   PACSTALL_DEBUG=1
   shift 1
@@ -804,15 +810,16 @@ function elevate() {
         [[ $NOCHECK ]] && nocheck="-Nc"
         ${PACSTALL_VERBOSE} || quiet="-Q"
         [[ $NOSANDBOX ]] && nosandbox="-Ns"
+        [[ -v PACSTALL_LOG ]] && xtrace="PACSTALL_LOG=${PACSTALL_LOG}"
 
-        trap - ERR RETURN
-        unset ignore_stack
         sudo PACSTALL_SUPPRESS_SOLUTIONS="$PACSTALL_SUPPRESS_SOLUTIONS" \
             PACSTALL_BUILD_CORES="$PACSTALL_BUILD_CORES" PACSTALL_EDITOR="$PACSTALL_EDITOR" \
             PACSTALL_DOWNLOADER="$PACSTALL_DOWNLOADER" PACSTALL_PAYLOAD="$PACSTALL_PAYLOAD" \
             PACSTALL_TMPDIR="$PACSTALL_TMPDIR" NO_COLOR="$NO_COLOR" SUDO_USER="${SUDO_USER}" \
-            DISABLE_PROMPTS="$DISABLE_PROMPTS" PACSTALL_ELEVATED=true \
+            DISABLE_PROMPTS="$DISABLE_PROMPTS" PACSTALL_ELEVATED=true $xtrace \
             pacstall $debug $keep $build $nocheck $quiet $nosandbox $@
+        cat "${PACSTALL_LOG}" "${PACSTALL_LOG}.elevated" > ${PACSTALL_LOG}
+        sudo rm "${PACSTALL_LOG}.elevated"
         exit $?
     fi
 }
--- a/pacstall
+++ b/pacstall
@@ -48,11 +48,13 @@ PACSTALL_DEBUG=0
 if [[ "${1}" == "-x" || "${1}" == "--debug" ]]; then
     if [[ -n ${PACSTALL_XTRACELOG} && -n ${PACSTALL_XTRACEFD} ]]; then
         if ! [[ ${PACSTALL_XTRACEFD} =~ ^[0-9]+$ ]]; then
-            # avoid having to duplicate gettext string used in later checks
-            printf '[\033[1;31m!\033[0m] \033[1mERROR\033[0m: '
-            printf $"'%s' must be an unsigned integer" 'PACSTALL_XTRACEFD' >&2
-            # avoid having the newline in a gettext string
-            printf '\n'
+            {
+                # avoid having to duplicate gettext string used in later checks
+                printf '[\033[1;31m!\033[0m] \033[1mERROR\033[0m: ';
+                printf $"'%s' must be an unsigned integer" 'PACSTALL_XTRACEFD';
+                # avoid having the newline in a gettext string
+                printf '\n';
+            } >&2
             exit 1
         fi
         # Separate elevated log file because of permissions issues
--- a/pacstall
+++ b/pacstall
@@ -46,20 +46,20 @@ export PACSTALL_INSTALL=1
 
 PACSTALL_DEBUG=0
 if [[ "${1}" == "-x" || "${1}" == "--debug" ]]; then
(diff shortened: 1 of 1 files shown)


pacstall
```diff
@@ -46,6 +46,12 @@ export PACSTALL_INSTALL=1
 
 PACSTALL_DEBUG=0
 if [[ "${1}" == "-x" || "${1}" == "--debug" ]]; then
+  if [[ -v PACSTALL_LOG ]]; then
+    # Separate elevated log file because of permissions issues
+    [[ $PACSTALL_ELEVATED ]] && PACSTALL_LOG="${PACSTALL_LOG}.elevated"
+    exec 3>>"${PACSTALL_LOG}"
+    export BASH_XTRACEFD=3
```
oklopfer: The 3 is arbitrary and a user choice, we need a separate variable for this

pacstall
```diff
@@ -46,6 +46,12 @@ export PACSTALL_INSTALL=1
 
 PACSTALL_DEBUG=0
 if [[ "${1}" == "-x" || "${1}" == "--debug" ]]; then
+  if [[ -v PACSTALL_LOG ]]; then
+    # Separate elevated log file because of permissions issues
+    [[ $PACSTALL_ELEVATED ]] && PACSTALL_LOG="${PACSTALL_LOG}.elevated"
```
oklopfer: If you are changing the var here it's going to eval as .elevated in the instance below when you cat

oklopfer (changes requested):

pacstall
oklopfer: Also why 3>> and not 3>?

pacstall
```diff
@@ -804,13 +810,16 @@ function elevate() {
...
+        [[ -v PACSTALL_LOG ]] && xtrace="PACSTALL_LOG=${PACSTALL_LOG}"
 
         sudo PACSTALL_SUPPRESS_SOLUTIONS="$PACSTALL_SUPPRESS_SOLUTIONS" \
             PACSTALL_BUILD_CORES="$PACSTALL_BUILD_CORES" PACSTALL_EDITOR="$PACSTALL_EDITOR" \
             PACSTALL_DOWNLOADER="$PACSTALL_DOWNLOADER" PACSTALL_PAYLOAD="$PACSTALL_PAYLOAD" \
             PACSTALL_TMPDIR="$PACSTALL_TMPDIR" NO_COLOR="$NO_COLOR" SUDO_USER="${SUDO_USER}" \
-            DISABLE_PROMPTS="$DISABLE_PROMPTS" PACSTALL_ELEVATED=true \
+            DISABLE_PROMPTS="$DISABLE_PROMPTS" PACSTALL_ELEVATED=true $xtrace \
             pacstall $debug $keep $build $nocheck $quiet $nosandbox $@
+        cat "${PACSTALL_LOG}" "${PACSTALL_LOG}.elevated" > ${PACSTALL_LOG}
```
oklopfer: Why not just cat the elevated log on top with >>? Less duplication

pacstall
```diff
@@ -46,6 +46,12 @@ export PACSTALL_INSTALL=1
 
 PACSTALL_DEBUG=0
 if [[ "${1}" == "-x" || "${1}" == "--debug" ]]; then
+  if [[ -v PACSTALL_LOG ]]; then
```
oklopfer: Should be -n and not -v pretty sure, the latter would open it up to more cases

oklopfer (changes requested):

pacstall
D-Brox: Typo

pacstall
oklopfer: As a result of switching to this, two things happened:
1. the end of lower PACSTALL_LOG now correctly prints to completion
2. however, there seems to be a minor race condition, which can be seen at the start of the elevated `def_colors` call (because it is so early, it might not be worth caring about)

oklopfer:
FYI this will need a new gettext string, but because #1458 is doing the same, we may as well add it in that one after this is merged

pacstall
```diff
@@ -45,12 +45,27 @@ export PACSTALL_USER=$(logname 2> /dev/null || echo "${SUDO_USER:-${USER:-$(whoa
 export PACSTALL_INSTALL=1
 
 PACSTALL_DEBUG=0
+if [[ -n ${PACSTALL_XTRACE} ]] && ! [[ ${PACSTALL_XTRACE} =~ ^[0-9]+$ ]]; then
+    printf $'%b must be an integer' '[\033[1;31m!\033[0m] \033[1mERROR\033[0m: PACSTALL_XTRACE' >&2
+    # avoid having the newline in a gettext string
+    printf '\n'
+    exit 1
+fi
```
D-Brox: Move this to inside the debug if statement

pacstall
```diff
@@ -810,8 +825,14 @@ function elevate() {
             PACSTALL_DOWNLOADER="$PACSTALL_DOWNLOADER" PACSTALL_PAYLOAD="$PACSTALL_PAYLOAD" \
             PACSTALL_TMPDIR="$PACSTALL_TMPDIR" NO_COLOR="$NO_COLOR" SUDO_USER="${SUDO_USER}" \
             DISABLE_PROMPTS="$DISABLE_PROMPTS" PACSTALL_ELEVATED=true \
+            PACSTALL_LOG="${PACSTALL_LOG}" PACSTALL_XTRACE="${PACSTALL_XTRACE}" \
             pacstall $debug $keep $build $nocheck $quiet $nosandbox $@
-        exit $?
+        exit_code=$?
+        if [[ -n ${debug} && -n ${PACSTALL_LOG} && -n ${PACSTALL_XTRACE} ]]; then
+            cat "${PACSTALL_LOG}.elevated" >> "${PACSTALL_LOG}" \
```
D-Brox: You can `set +x` right before the cat to avoid the race condition

oklopfer: Wouldn't that just go back to the problem of not getting the full debug output from the lower run?

pacstall
oklopfer: True

pacstall
oklopfer: I guess it's fine since `exit_code` is shown before and all it does after is cat and rm

oklopfer (dismissed):

oklopfer:
merge if you think it looks good @D-Brox, all seems to work as expected

oklopfer (dismissed):

pacstall
```diff
@@ -46,11 +46,28 @@ export PACSTALL_INSTALL=1
 
 PACSTALL_DEBUG=0
 if [[ "${1}" == "-x" || "${1}" == "--debug" ]]; then
-  set -x
-  PACSTALL_DEBUG=1
-  shift 1
+    if [[ -n ${PACSTALL_XTRACELOG} && -n ${PACSTALL_XTRACEFD} ]]; then
+        if ! [[ ${PACSTALL_XTRACEFD} =~ ^[0-9]+$ ]]; then
+            # avoid having to duplicate gettext string used in later checks
+            printf '[\033[1;31m!\033[0m] \033[1mERROR\033[0m: '
```
Elsie19:
```suggestion
            printf '[\033[1;31m!\033[0m] \033[1mERROR\033[0m: ' >&2
```

Or do you want `stderr` to be clean?

Elsie19 (changes requested):

oklopfer (approved):

PR #1470, opened 2026-05-20
docs(readme): add package count badge

aKqir24:
## Purpose

To better monitor the number of packages in the `pacstall-programs` repo.

## Approach

I used a shields badge that counts the repo lines of code in pacstall-programs/packagelist containing all the pacscripts. It is not that big of a change, and I think it is not really necessary to put this, if you like it or not let me know:)

## Progress

- [x] Put the badge
- [x] Change to read from packagelist


oklopfer:
There are more packages than files, actually: https://github.com/pacstall/pacstall-programs/blob/master/packagelist

Elsie19:
Can this be done doing a line count of a file or no?

aKqir24:
> There are more packages than files, actually: https://github.com/pacstall/pacstall-programs/blob/master/packagelist

I see, thanks for looking it up, let me see if I can do something

aKqir24:
> Can this be done doing a line count of a file or no?

I changed it to [this](https://github.com/pajecawav/ghloc-web), it will filter to count in packagelist, it took in the packages repo.

oklopfer (approved):
Looks great!

PR #1476, opened 2026-06-15
fix(bwrap): use full env context to filter new vars

D-Brox:
## Purpose

Not every var was being filtered correctly

## Approach

Source `bwrapenv` before getting the old env.

## Progress

- [x] Do it
- [x] Test it (thanks @oklopfer) 

## Checklist


oklopfer (approved):
Works

PR #1480, opened 2026-06-18
fix(pacstall,package-base.sh): PACDIR should never be a file

### LINKED ISSUE #1471: /tmp/pacstall is not a folder
### Describe the bug

pacstall create /tmp/pacstall as a file but not a folder

pacstall version Coral

### To reproduce

pacstall -I yourscript.pacstall

Can happen but not always. always in distrobox, not always on debian 12 VM.

quick fix : sudo rm /tmp/pacstall, mkdir -p /tmp/pacstall

### Expected behavior

pacstall should create the folder it need

### Screenshots

Image

### OS version

Debian 

### Pacstall version

4.6.0

oklopfer:
## Purpose

PACDIR might accidentally be created as a file instead of a folder

## Approach

Ensure creation of PACDIR before copying, ensure removal of a PACDIR file at init if exists

## Progress

- [x] do it
- [x] test it

## Addendum

Closes #1471.

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -696,6 +696,10 @@ if ! $(cd "${PACTMP}" 2> /dev/null); then
     exit 1
 fi
 
+if [[ -f "${PACDIR}" ]]; then
+    rm -rf "${PACDIR}"
+fi
+
 if [[ ! -t 0 ]]; then
     NON_INTERACTIVE=true
     fancy_message warn $"Reading input from pipe"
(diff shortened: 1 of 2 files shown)


pacstall
```diff
@@ -696,6 +696,10 @@ if ! $(cd "${PACTMP}" 2> /dev/null); then
     exit 1
 fi
 
+if [[ -f "${PACDIR}" ]]; then
+    rm -rf "${PACDIR}"
```
oklopfer: should confirm we don't need sudo here

Elsie19: If the check succeeds, it's not a file so,
```suggestion
    rm -f "${PACDIR}"
```

Elsie19 (changes requested):

Elsie19 (approved):

PR #1491, opened 2026-07-30
chan(pacstall): fragile word split ~> mapfile / chan(pacstall): use extglob for flag parsing

Elsie19:
## Purpose

Replaces some fragile `arr=($(cmd))` with `mapfile` and also reduce code size by using `extglob` so we can basically regex match flags instead of "*if or*ing" a bunch.

## Addendum

>[!IMPORTANT]
> Make sure to merge this, not squash merge!

## Checklist

### DIFF
--- a/pacstall
+++ b/pacstall
@@ -659,7 +659,7 @@ function getMasks() {
 
 function getMasks_offending_pkg() {
     { ignore_stack=false; set -o pipefail; trap stacktrace ERR RETURN; }
-    local the_name="${1}"
+    local the_name="${1}" pkgs
     mapfile -t pkgs < <(pacstall -L)
     if ((${#pkgs[@]} == 0)); then
         return 0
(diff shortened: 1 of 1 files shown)


pacstall
```diff
@@ -794,14 +801,14 @@ function lock() {
...
-        if [[ ${#instances[@]} -lt $valid ]]; then
+        mapfile -t instances < <(pidof -o %PPID -x "$0" -d $'\n')
+        if (( ${#instances[@]} < valid )); then
             { ignore_stack=true; return 1; }
         fi
         return 0
     fi
 
-    pidof -o %PPID -x "$0" > /dev/null && return 0 || { ignore_stack=true; return 1; }
+    pidof -q -o %PPID -x "$0" && return 0 || { ignore_stack=true; return 1; }
```
D-Brox: Do this on #1488 instead

D-Brox (changes requested):

D-Brox (approved):

PR #1496, opened 2026-08-20
feat(incompatible): add optional failure reason

Elsie19:
## Purpose

You can now add an optional message to `incompatible` that tells the user why exactly a certain distro clamp won't work:

```bash
incompatible=("debian:sid[package ships rm -rf on Sid]" "ubuntu:*[I just hate Ubuntu]")
```

## Checklist

### DIFF
--- a/misc/scripts/fetch-sources.sh
+++ b/misc/scripts/fetch-sources.sh
@@ -586,7 +586,7 @@ function set_distro() {
 function get_compatible_releases() {
     { ignore_stack=false; set -o pipefail; trap stacktrace ERR RETURN; }
     # example for this function is "ubuntu:jammy"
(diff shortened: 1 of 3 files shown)


misc/scripts/fetch-sources.sh
```diff
@@ -645,25 +649,30 @@ function get_incompatible_releases() {
             fi
         fi
     fi
-    for key in "${incomp_list[@]}"; do
+    for incompat in "${incomp_list[@]}"; do
+        key="${incompat,,}"
+        key="${key%%\[*}"
+        reason="${incompat#*\[}"
+        [[ $reason != "${incompat}" ]] && reason="${reason%\]}"
```
oklopfer: Pls turn this one into an if check otherwise stacktrace will read a failure and break methinks

oklopfer (changes requested):

oklopfer (approved):

PR #1505, opened 2026-09-20
fix(package.sh): don't create metadata if only building

oklopfer:
## Purpose
We are erroneously creating metadata even when not installing

## Approach
Skip it when -B flag is active

## Progress
- [x] Do
- [x] Test

## Checklist


Elsie19 (approved):
probably works lgtm

